PR #1, opened 2016-08-24
Started the registrar.sol file

alexvandesande:
Testings in progress


DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
+// Name registrar using deposits
+
+contract Deed {
+    // The Deed is a contract intended simply to hold ether
+    // It can be controlled only by the registrar and can only send ether back to the owner
+    address public owner;
```
Arachnid: Would it be possible to just make 'owner' and 'bidder' the same, to simplify matters?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    address public bidder;
+    Registrar public registrar;
+    bytes32 public sealedBid;
+    bytes32 public hashedName;
```
Arachnid: In ENS, I've been calling this 'label' - whatever we call it, we should be consistent.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    uint public value;
+    address public burn = 0xdead;
```
Arachnid: Should be flagged `constant`

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    
+    function Deed(address _bidder, bytes32 _sealedBid, address _registrar) {
+        bidder = _bidder;
+        registrar = Registrar(_registrar);
+        sealedBid = _sealedBid;
+        creationDate = now;
+    }
+    
+    function unsealBid(bytes32 _hashedName, address _owner, uint _value, bytes32 _salt) noEther onlyRegistrar {
+        if (sha3(_hashedName, _owner, _value, _salt) != sealedBid) throw;
```
Arachnid: This sha3 operation is also performed by the Registrar; could we abstract it into a (ridiculously simple) library?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        hashedName = _hashedName;
+        owner = _owner;
+        value = _value;
+        if (!bidder.send(this.balance - _value)) throw;
+    }
+    
+    function closeDeed( uint burnAmount) noEther onlyRegistrar {
```
Arachnid: `burnAmount` should probably be called `burnFraction` or similar.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        burn.send((burnAmount * this.balance)/1000);
+        selfdestruct(bidder);
```
Arachnid: Could we avoid sending funds by selfdestruct? Doing so doesn't call fallback functions, so people with wallets won't see the funds arrive.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    
+    function transferOwnership(address newOwner) {
+        if (msg.sender != owner ) throw;
+        owner = newOwner;
+    }
+}
+
+contract Registrar {
+    mapping (bytes32 => entry) public hashedNames;
+    mapping (bytes32 => address) public sealedBids;
```
Arachnid: Any reason this couldn't be a `mapping (bytes32 => Deed)`?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    
+    enum Mode { Open, Auction, Owned }
+    uint auctionLength = 20 minutes;
```
Arachnid: These should all be `constant`

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    uint revealPeriod = 5 minutes;
+    uint renewalPeriod = 2 hours;
+    uint M = 1000000; // just a multiplier to get more precision on averages
+    uint public averagePeriod;
+    uint public lastSinceNewName;
+    uint public minBid = 0.1 ether;    // TODO: figure out a better way to come up with this
```
Arachnid: `msg.gas * 100000000` seems a reasonable approach, though we might want to use a moving average like the other values for it.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    uint public averagePrice = minBid;
+    
+    struct entry {
```
Arachnid: Can some of these be smaller than 256 bits, to save on storage costs?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        Mode status;
+        address deed;
+        uint auctionDate;
```
Arachnid: Given this is when the auction ends, not begins, should it be called something like `auctionExpiry`?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        uint value;
+        uint highestBid;
+        uint firstRegistered;
+        uint lastRenewed;
+        uint renewalDate;
+        uint averagePrice;
+    }
+    
+    // doesnt work
```
Arachnid: Should probably be deleted, then. :)

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        _        
+    }
+    
+    modifier noEther  {
+        if (msg.value > 0) throw;
+        _
+    }   
+    
+    function Registrar() noEther  {
+        lastSinceNewName = now;
```
Arachnid: This should probably initialise averagePeriod to something sensible, too.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    /*
+    ## Start Auction for available name
+
+    Anyone can start an auction by sending an array of hashes that they want to bid for. 
+    Arrays are sent so that someone can open up an auction for X dummy hashes when they 
+    are only really interested in bidding for one. This will increase the cost for an 
+    attacker from simply bidding on all new auctions blindly. Dummy auctions that are 
+    open but not bid on are closed after a week. 
+    */    
+    function openAuctions(bytes32[] auctionsArray) noEther{
```
Arachnid: Is it necessary for this to take an array? If it only takes one parameter, it'd be easier to call, and you can still call it multiple times.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+
+        for (uint i = 0; i < auctionsArray.length; i ++ ) {
+            entry newAuction = hashedNames[auctionsArray[i]];
+            if (newAuction.status == Mode.Owned && now < newAuction.renewalDate) {
```
Arachnid: This would be easier to follow if the top-level if statements acted as a switch on `newAuction.status`.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    Anyone can start an auction by sending an array of hashes that they want to bid for. 
+    Arrays are sent so that someone can open up an auction for X dummy hashes when they 
+    are only really interested in bidding for one. This will increase the cost for an 
+    attacker from simply bidding on all new auctions blindly. Dummy auctions that are 
+    open but not bid on are closed after a week. 
+    */    
+    function openAuctions(bytes32[] auctionsArray) noEther{
+
+        for (uint i = 0; i < auctionsArray.length; i ++ ) {
+            entry newAuction = hashedNames[auctionsArray[i]];
```
Arachnid: Could we call this `entry` or `registration`? It's not yet clear that it _is_ a new auction, which makes some of the following code a bit hard to understand.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+            if (newAuction.status == Mode.Owned && now < newAuction.renewalDate) {
+                throw; 
+            } else if (newAuction.status == Mode.Auction && now < newAuction.auctionDate) {
+                throw;
+            } else if (newAuction.status == Mode.Owned) {
+                Deed deedContract = Deed(newAuction.deed);
+                deedContract.closeDeed(1);
+            }
+            newAuction.auctionDate = now + auctionLength;
+            newAuction.renewalDate = now + renewalPeriod;
```
Arachnid: Shouldn't the registration period start when the auction ends?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    renewed, and the price paid for it. **The price paid is also saved on a moving average amount, 
+    calculated as ```averagePrice = averagePrice * 0.999 + newPrice * 0.001```**. The 
+    averagePrice at the moment of purchase is also registered on the contract.
+    */ 
+    function revealBid(bytes32 _hashedName, address _owner, uint _value, bytes32 _salt) noEther  {
+        bytes32 seal = shaBid(_hashedName, _owner, _value, _salt);
+        address bidAddress = sealedBids[seal];
+        if (bidAddress == 0 ) throw;
+        Deed bid = Deed(bidAddress);
+        bid.unsealBid(_hashedName, _owner, _value, _salt);
```
Arachnid: Is it necessary for the deed to unseal itself? If `owner` and `bidder` can be conflated, it doesn't need any information from the sealed bid; it just needs to be told to reduce its stored deposit by some amount.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        entry name = hashedNames[bid.hashedName()];
+        
+        if (bid.creationDate() > name.auctionDate - revealPeriod
```
Arachnid: Is there a reason to punish this behaviour so harshly?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+            bid.closeDeed(1);
+        }
+    }
+    
+    function finalizeAuction(bytes32 nameHash) noEther {
+        entry name = hashedNames[nameHash];
+        if (now < name.auctionDate || name.status != Mode.Auction) throw;
+        
+        // set the name
+        name.status = Mode.Owned;
```
Arachnid: What if nobody bid?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        name.firstRegistered = now;
+        name.lastRenewed = now;
+
+        //set the averages
+        uint period = (now - lastSinceNewName) * M;
+        averagePeriod = (9999 * averagePeriod + period) / 10000;
+        uint n = averagePeriod < (60 seconds * M) ? 7 days * M / averagePeriod : 10000;
```
Arachnid: What's happening here - how `60 seconds` and `7 days` are determined is a little unclear; could you add a comment?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        averagePrice = ((n-1)*averagePrice + name.value ) / n;
```
Arachnid: It's probably worth adding an `updateFilter` function for these calculations.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    after 2 years, 7/8 after 3 years etc). 
+    */ 
+    function updatedValue(bytes32 nameHash) constant returns (uint updatedPrice) {
+        entry name = hashedNames[nameHash];
+        return name.value * averagePrice / name.averagePrice;
+    }
+    
+    function renewDeed(bytes32 nameHash) {
+        entry name = hashedNames[nameHash];
+        if (now < name.firstRegistered + renewalPeriod/2 ) throw;
```
Arachnid: Comment says 'at any moment', but this restricts renewals to after half the period has elapsed.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        Deed deedContract = Deed(name.deed);
+        uint difference = 0;
+        if (msg.sender != deedContract.owner() && name.status != Mode.Owned) throw;
```
Arachnid: ||, not &&

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        } else {
+            difference =  name.value - updatedPrice; 
+            uint ratioOfRecovery = 2**((now - name.lastRenewed)/ 1 years);
+            // the more time, the more you recover more from your initial investment
+            deedContract.setBalance(name.value + difference - difference/ratioOfRecovery);
+        }
+        
+        name.value = updatedPrice;
+        name.lastRenewed = now;
+        // next renewal data is twice the current age
```
Arachnid: Personally, I think a fixed 1 yr renewal period would make more sense.

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    your ether. If there are better opportunities, like staking it, lending it or investing 
+    in some other new venture, then holding names should be seen as an undesirable outcome 
+    and owners have an incentive to release them. Names can be sold at any moment, but the 
+    buyer will incur the same renewal cost/benefit analysis. 
+    */    
+    function releaseDeed(bytes32 nameHash) noEther  {
+        entry name = hashedNames[nameHash];
+        Deed deedContract = Deed(name.deed);
+        if (now < name.firstRegistered + renewalPeriod ) throw;
+        if (msg.sender != deedContract.owner() && name.status != Mode.Owned) throw;
```
Arachnid: ||, not &&

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+    The real cost of holding a name is the opportunity cost of doing something better with 
+    your ether. If there are better opportunities, like staking it, lending it or investing 
+    in some other new venture, then holding names should be seen as an undesirable outcome 
+    and owners have an incentive to release them. Names can be sold at any moment, but the 
+    buyer will incur the same renewal cost/benefit analysis. 
+    */    
+    function releaseDeed(bytes32 nameHash) noEther  {
+        entry name = hashedNames[nameHash];
+        Deed deedContract = Deed(name.deed);
+        if (now < name.firstRegistered + renewalPeriod ) throw;
```
Arachnid: Why this restriction?

DepositRegistrar.sol
```diff
@@ -0,0 +1,306 @@
...
+        if (msg.sender != deedContract.owner() && name.status != Mode.Owned) throw;
+        
+        name.status = Mode.Open;
+        deedContract.closeDeed(0);
+    }
+}
+
```
Arachnid: Lots of whitespace here.

Arachnid:
One major thing missing is actually integrating with ENS, but I assume you intended to leave that for a future PR?

DepositRegistrar.sol
alexvandesande: During the bid auction separating both allows greater anonymity. But maybe bidder is unnecessary to keep.

DepositRegistrar.sol
alexvandesande: Yeah, naming things has been hard. This is how I went to it:

Name: the thing we are bidding and own
Hashed Name: the proper name, passed through a sha3 to add anonymity
Deed: the contract that holds information about the property. I believe Deed is a good term

I don't understand what "label" means in this context

DepositRegistrar.sol
Arachnid: Fair enough. Someone desiring that would have to be able to send the bid from a contract that can't be traced back to them too, though, so I'm not sure the gain in anonymity is worth the complexity.

DepositRegistrar.sol
Arachnid: So, a label is a part of a name - in 'foo.bar.com', 'foo', 'bar', and 'com' are labels. In ENS I've been using label to refer to the hash of the actual label, since raw labels aren't handled anywhere in ENS.

Since a registrar will only register names on top of a single TLD, passing in the whole name seems unnecessary - just the one component being registered will suffice, and ensure people don't try and register names that the registrar isn't authoritative for.

DepositRegistrar.sol
alexvandesande: mmm didn't know that! I can reverse the process, I send funds to the owner and then burn by selfdetruct.

I wanted to do this way to avoid any funny business with using a fallback function, but we can try to be extra careful;

DepositRegistrar.sol
alexvandesande: Problem with that approach is that miners can create transactions with a very low msg.gas, so someone could pay a miner to send a lot of 0 transactions and move the average way down

DepositRegistrar.sol
Arachnid: Yup, understood. You definitely shouldn't `throw` if `msg.send` fails in this case, since otherwise a malicious user could prevent someone outbidding them with a fallback function that throws!

DepositRegistrar.sol
alexvandesande: But I want it to work! Any help? 😞

DepositRegistrar.sol
alexvandesande: At some point I was doing it but felt it wouldn't really matter after a few thousand transactions

DepositRegistrar.sol
Arachnid: Hm, I'd considered that someone could create TXes with low gas, but couldn't think of a way to reliably do so - you're right, miners would be one way to do that.

I believe there are plans for a 'gas economy' that would eliminate this attack by making gas price part of consensus, but that's not here yet.

A couple of options:
 - 1/nth of the current average price, for some n
 - Depend on an external 'gas price oracle' (ugh)

DepositRegistrar.sol
Arachnid: You can pass parameters _into_ modifiers, but I don't think there's any way to pass them out. I think the way you've got it is the best way available at the moment.

DepositRegistrar.sol
alexvandesande: I would like the name registrar Dapp to randomly initiate some names whenever someone opens an auction, without requiring extra calls from the user. This guarantees more privacy by default.

DepositRegistrar.sol
alexvandesande: This was necessary before, but now after some rewrite I don't see why not remove it

DepositRegistrar.sol
alexvandesande: The bid requires the owner variable to execute transferOwnership. Bidders and owners should not be the same, to allow more anonymity during the auctions.

Arachnid: I'm still skeptical separating the two adds anonymity; the owner still has to get funds into a bidder address somehow, in which case they may as well just make that account the owner, and transfer it afterwards.

DepositRegistrar.sol
Arachnid: Could we offer a version that takes a single argument, and call it in a loop from this one, then?

DepositRegistrar.sol
alexvandesande: Yes. Otherwise you can make a bid and choose not to reveal during the reveal period, after some others have revealed. This would change the auction from a blind vickrey into a completely different auction. As an example of what someone could do, they could bid multiple values for any auction and then blackmailing the current winner into paying them NOT to reveal a bid that would be low enough not to win, but high enough to increase the ask price of the name.

DepositRegistrar.sol
alexvandesande: If nobody bid then the auction can be opened again at the newAuction() function. But I could manually deactivate it here too.

DepositRegistrar.sol
Arachnid: Good point. Shouldn't we just take 100%, then?

DepositRegistrar.sol
alexvandesande: It checks if there were, in average more than 10,000 registration per week (averagePeriod less than 60.48 seconds). If so, use the number of average registrations, elsewhere, use 10,000. I'm using period as a simpler way to measure frequency.

DepositRegistrar.sol
alexvandesande: updateFilter?

alexvandesande: Ah you meant the ability to change the equation?

DepositRegistrar.sol
alexvandesande: Yes. Will change the description. Otherwise someone could register dummy names at almost no cost to manipulate the averages.

DepositRegistrar.sol
Arachnid: Won't it be owned, with a renewal date in the future, and thus unclaimable until then?

DepositRegistrar.sol
Arachnid: Now I know that; could you add a comment that a period of 60 seconds == 10k registrations per week so others do? :)

DepositRegistrar.sol
Arachnid: I meant that there's two different filters here, both of the form `((a-b) * old + new) / a`; refactoring them into a function that updates and returns the filter value might make that easier to follow.

DepositRegistrar.sol
alexvandesande: It's && because you shouldn't be able to renew a deed that is not currently being owned by anyone.

It's controversial, to allow someone to renew a name they don't own: depending on the date you choose to renew it, it will affect the amount you'll need to pay (or receive) on the next renewal. Not sure if someone else should be allowed to do it..

Arachnid: With &&, this will only throw if you're both not the owner and it's not owned, though. If the name is owned by someone else, this will evaluate to (false && false), and fail to throw.

DepositRegistrar.sol
alexvandesande: one year goes rather quickly. I hate to renew domains every year, specially if it's a hassle and you own many. This way you have to renew after 1 year, then after 2 years, 4 etc. Maybe we can add a upper limit, someone only needing to renew it after 8 or 16 years might be a stretch, but hey, maybe when ethereum has existed as long as the web, it might make sense.

DepositRegistrar.sol
alexvandesande: Ah just noticed. You're right the restriction shouldn't apply here

DepositRegistrar.sol
alexvandesande: This is where the restriction should be added to half the renewalPeriod. Otherwise someone could register dummy names at almost no cost to manipulate the averages.

DepositRegistrar.sol
alexvandesande: I hate having to look at the bottom of the screen while coding, so I always add lots of whitespace at the end. No reason to have in on the github.

alexvandesande:
I was hoping you'd help me with the proper integration!

DepositRegistrar.sol
Arachnid: Fair enough; I think we should definitely put an upper limit on it, yes.

Arachnid:
Very happy to. It's pretty straightforward.

DepositRegistrar.sol
alexvandesande: Well then I'm using name to refer to what you call a second level label: foo.eth.
I could rename it all to label here for consistency, but are you sure this is the proper terminology? I don't think they are very self explanatory. Maybe if we all used "domains"?

DepositRegistrar.sol
alexvandesande: Why not simply call with a single item array?

DepositRegistrar.sol
alexvandesande: You can hire a service to bid for you and do off chain transactions..?

DepositRegistrar.sol
alexvandesande: I suppose getting something back would make someone reveal it?

DepositRegistrar.sol
Arachnid: If you just have users submit `sha3('foo')` instead of `sha3('foo.eth')`, then you can pass this hashed token directly to the ENS registrar; no need to deal with unhashed names at all.

In DNS terminology, `foo.bar` is a domain, while `foo` and `bar` are labels; there's no convention for `H('foo')`, but I've just been calling them labels, since we don't use the unhashed ones anywhere.

DepositRegistrar.sol
Arachnid: I'm just thinking about ease-of-use from other contracts, and from UIs like Mist; it's easier to submit a single hash than an array of them, and it's easier to read the code without the loop around it.

DepositRegistrar.sol
Arachnid: Well, then that service could also handle transferring the name to you after you win. :)

DepositRegistrar.sol
Arachnid: Good point. Perhaps they should get more than 0.1%, then - 1%? 10%?

HashRegistrar.sol
```diff
@@ -0,0 +1,282 @@
...
+        registrar = Registrar(_registrar);
+        sealedBid = _sealedBid;
+        creationDate = now;
+    }
+    
+    function unsealBid(bytes32 _hash, address _owner, uint _value, bytes32 _salt) noEther onlyRegistrar {
+        if (sha3(_hash, _owner, _value, _salt) != sealedBid) throw;
+        hash = _hash;
+        owner = _owner;
+        if (!owner.send(this.balance - _value)) throw;
```
Arachnid: Should this come before the previous line, in case it updates the owner?

HashRegistrar.sol
```diff
@@ -0,0 +1,282 @@
...
+    }
+    
+    function transferOwnership(address newOwner) {
+        if (msg.sender != owner ) throw;
+        owner = newOwner;
+    }
+}
+
+contract Registrar {
+    mapping (bytes32 => entry) public hashes;
```
Arachnid: Maybe call this `entries` instead?

HashRegistrar.sol
```diff
@@ -0,0 +1,282 @@
...
+    uint32 constant revealPeriod = 5 minutes;
+    uint32 constant renewalPeriod = 2 hours;
+    uint32 constant M = 1000000; // just a multiplier to get more precision on averages
+    uint public averagePeriod;
+    uint public lastSinceNewRegistry;
+    uint public averagePrice;
+    
+    struct entry {
+        Mode status;
+        address deed;
```
Arachnid: `Deed deed` instead?

HashRegistrar.sol
```diff
@@ -0,0 +1,282 @@
...
+    }
+    
+    function Deed(bytes32 _sealedBid, address _registrar) {
+        registrar = Registrar(_registrar);
+        sealedBid = _sealedBid;
+        creationDate = now;
+    }
+    
+    function unsealBid(bytes32 _hash, address _owner, uint _value, bytes32 _salt) noEther onlyRegistrar {
+        if (sha3(_hash, _owner, _value, _salt) != sealedBid) throw;
```
Arachnid: Can this be modified to only run if _hash is 0? That way, it's guaranteed it can only run once, and therefore there's no way the original owner can steal it back by 'revealing' their bid again. I know this is `onlyRegistrar`, but the added self-contained security wouldn't hurt.

HashRegistrar.sol
```diff
@@ -0,0 +1,282 @@
...
+    ## Start Auction for available hash
+
+    Anyone can start an auction by sending an array of hashes that they want to bid for. 
+    Arrays are sent so that someone can open up an auction for X dummy hashes when they 
+    are only really interested in bidding for one. This will increase the cost for an 
+    attacker from simply bidding on all new auctions blindly. Dummy auctions that are 
+    open but not bid on are closed after a week. 
+    */    
+    function startAuction(bytes32 _hash) noEther{
+        entry hash = hashes[_hash];
```
Arachnid: This is a bit confusing, since `_hash` is a hash, but `hash` is an `entry`.

HashRegistrar.sol
```diff
@@ -0,0 +1,282 @@
...
+            startAuction(_hashes[i]);
+        }
+    }
+    
+    function shaBid(bytes32 hash, address owner, uint value, bytes32 salt) constant returns (bytes32 sealedBid) {
+        return sha3(hash, owner, value, salt);
+    }
+    
+    /*
+    ## One week auction for the desired hash
```
Arachnid: Probably shouldn't say "one week"

HashRegistrar.sol
alexvandesande: Or rename entry as hash? Not sure which would make the code clearer. Maybe the word "hash" is already being overused but at least it's more specific than entry..

HashRegistrar.sol
alexvandesande: I believe the condition requiring hash to be 0 takes care of this problem. I think having the send as the last thing is always a good practice

HashRegistrar.sol
```diff
@@ -0,0 +1,282 @@
...
+    contains basic information like when the hash was first registered, when it needs to be 
+    renewed, and the price paid for it. **The price paid is also saved on a moving average amount, 
+    calculated as ```averagePrice = averagePrice * 0.999 + newPrice * 0.001```**. The 
+    averagePrice at the moment of purchase is also registered on the contract.
+    */ 
+    function revealBid(bytes32 _hash, address _owner, uint _value, bytes32 _salt) noEther  {
+        bytes32 seal = shaBid(_hash, _owner, _value, _salt);
+        Deed bid = sealedBids[seal];
+        if (address(bid) == 0 ) throw;
+        bid.unsealBid(_hash, _owner, _value, _salt);
```
Arachnid: Thinking about it more, could this be skipped entirely? Instead, the Deed could start without an owner, and the registrar could be permitted to call and set the owner one time only. It can then call setBalance to refund any extra balance.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+        sealedBids[seal] = Deed(0);
+        
+        entry hash = hashes[bid.hash()];
+        
+        if (bid.creationDate() > hash.auctionExpiration - revealPeriod
+            || now > hash.auctionExpiration ) {
+            // bid is invalid, refund 99.9%
+            bid.closeDeed(1);
+            
+        } else if ( _value < averagePrice / 100 ) {
```
Arachnid: This should probably be a parameter.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+            // bid is invalid but not punishable
+            bid.closeDeed(1000);
+            
+        } else if (_value > hash.highestBid) {
+            // new winner
```
Arachnid: This section needs to remove the deed from `sealedBids`, or the winner can wait until the auction is over, then 'reveal' their bid again and get a refund in the first clause of this if statement.

HashRegistrar.sol
alexvandesande: Mmm I don't see why not. It would indeed simplify the code, which is always a good thing

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+            hash.value = _value;
+            bid.closeDeed(999);
+            
+        } else {
+            // bid doesn't affect auction
+            bid.closeDeed(999);
+        }
+    }
+    
+    function cancelBid(bytes32 seal) {
```
Arachnid: I thought we didn't want to permit getting deposits back without revealing a valid bid, to prevent blackmail etc?

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+        averagePrice = (n*averagePrice + hash.value ) / (n+1);
+        hash.averagePrice = averagePrice;
+        lastSinceNewRegistry = now;
+        
+        Deed deedContract = Deed(hash.deed);
+        deedContract.setBalance(hash.value);
+    }
+    
+    /*
+    ## Renewals can be done at any moment by renewing the deposit
```
Arachnid: Still at any moment? :)

Arachnid: I actually agreed with your reasoning that allowing people to unregister almost immediately made it too easy to submit bids to escalate prices.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+    you have the right to withdraw part of the difference (1/2 after a year, 3/4 
+    after 2 years, 7/8 after 3 years etc). 
+    */ 
+    function updatedValue(bytes32 _hash) constant returns (uint updatedPrice) {
+        entry hash = hashes[_hash];
+        return hash.value * averagePrice / hash.averagePrice;
+    }
+    
+    function renewDeed(bytes32 _hash) {
+        entry hash = hashes[_hash];
```
Arachnid: I actually think you were right in your reasoning; this should be limited to the owner only.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+            difference =  hash.value - updatedPrice; 
+            uint ratioOfRecovery = 2**((now - hash.lastRenewed)/ renewalPeriod);
+            // the more time, the more you recover more from your initial investment
+            deedContract.setBalance(hash.value + difference - difference/ratioOfRecovery);
+        }
+        
+        hash.value = updatedPrice;
+        hash.lastRenewed = now;
+        // next renewal data is twice the current age
+        hash.renewalDate = 2 * now - hash.firstRegistered;
```
Arachnid: I do think this needs a cap of, say, 4 years.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+    }
+     
+    /*
+    ## Renewals can be done at any moment by renewing the deposit
```
Arachnid: Renewals?

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+The contract is called hash registrar because it deals with onwership of hashes of 
+things, never the things themselves, to increase privacy and extensibility.
+
+*/
+
+contract Deed {
+    // The Deed is a contract intended simply to hold ether
+    // It can be controlled only by the registrar and can only send ether back to the owner
+    Registrar public registrar;
+    bytes32 public sealedBid;
```
Arachnid: This isn't required any longer.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+    renewed, and the price paid for it. **The price paid is also saved on a moving average amount, 
+    calculated as ```averagePrice = averagePrice * 0.999 + newPrice * 0.001```**. The 
+    averagePrice at the moment of purchase is also registered on the contract.
+    */ 
+    function revealBid(bytes32 _hash, address _owner, uint _value, bytes32 _salt) noEther  {
+        bytes32 seal = shaBid(_hash, _owner, _value, _salt);
+        Deed bid = sealedBids[seal];
+        if (address(bid) == 0 ) throw;
+        bid.setOwner(_owner);
+        bid.setBalance(_value);
```
Arachnid: You could save a small amount of gas by only doing this in the one branch below in which the deed isn't immediately closed.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+        //Calculate the moving average period as a way to measure frequency
+        uint period = (now - lastSinceNewRegistry) * M;
+        averagePeriod = (999 * averagePeriod + period) / 1000;
+        // 60 seconds is the average of 10,000 registrations per week. Use whatever's higher.
+        uint n = averagePeriod < (60 seconds * M) ? 7 days * M / averagePeriod : 9999;
+        averagePrice = (n*averagePrice + h.value ) / (n+1);
+        h.averagePrice = averagePrice;
+        lastSinceNewRegistry = now;
+        
+        Deed deedContract = Deed(h.deed);
```
Arachnid: This cast is now superfluous.

HashRegistrar.sol
```diff
@@ -0,0 +1,320 @@
...
+    The real cost of holding a hash is the opportunity cost of doing something better with 
+    your ether. If there are better opportunities, like staking it, lending it or investing 
+    in some other new venture, then holding hashes should be seen as an undesirable outcome 
+    and owners have an incentive to release them. Hashes can be sold at any moment, but the 
+    buyer will incur the same renewal cost/benefit analysis. 
+    */  
+    
+    function renewDeed(bytes32 _hash) {
+        entry h = entries[_hash];
+        Deed deedContract = Deed(h.deed);
```
Arachnid: This cast is now superfluous.

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+    
+    function setBalance(uint newValue) onlyRegistrar {
+        // Check if it has enough balance to set the value
+        if (this.balance < newValue) throw;
+        // Send the difference to the owner
+        if (!owner.send(this.balance - newValue)) throw;
+    }
+
+    function closeDeed(uint refundRatio) onlyRegistrar {
+        owner.send((refundRatio * this.balance)/1000);
```
chriseth: Please check the return value.

HashRegistrar.sol
chriseth: You should actually never explicitly convert a contract type.

HashRegistrar.sol
Arachnid: Actually, this is deliberate: otherwise the owner could prevent someone from outbidding them by throwing in their fallback function.

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+require the owner to update their deposit to what is assumed to be the market 
+rate of the name of their current name. If names in general have increased in 
+value but the owner is among the bottom half that hasn't done anything to increase 
+the value of their own names, then all possible profit will be negated by the extra 
+deposit. If names have decreased in value the owners can request to withdraw part 
+of the deposit and the cost of the name will be the cost of opportunity of possibly 
+having invested in something that would have a better return. At any point name 
+holders can release and get the full deposit back.
+
+The contract is called hash registrar because it deals with onwership of hashes of
```
Smithgift: s/onwership/ownership/

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+    /*
+    ## Start Auction for available hash
+
+    Anyone can start an auction by sending an array of hashes that they want to bid for. 
+    Arrays are sent so that someone can open up an auction for X dummy hashes when they 
+    are only really interested in bidding for one. This will increase the cost for an 
+    attacker from simply bidding on all new auctions blindly. Dummy auctions that are 
+    open but not bid on are closed after a week. 
+    */    
+    function startAuction(bytes32 _hash) noEther{
```
Smithgift: `noEther{` should probably be `noEther {`

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+            Deed deedContract = Deed(newAuction.deed);
+            deedContract.closeDeed(999);
+        }
+        
+        newAuction.auctionExpiration = now + auctionLength;
+        newAuction.status = Mode.Auction;        
+    }
+
+    // Allows you to open multiple for better anonimity
+    function startAuctions(bytes32[] _hashes) noEther{
```
Smithgift: And here.

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+        uint auctionExpiration;
+        uint value;
+        uint highestBid;
+        uint firstRegistered;
+        uint lastRenewed;
+        uint renewalDate;
+        uint averagePrice;
+    }
+    
+    modifier noEther  {
```
Smithgift: Extra space between noEther and {

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+        if (msg.value > 0) throw;
+        _
+    }   
+    
+    function Registrar() noEther  {
```
Smithgift: Another extra space

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+        return sha3(hash, owner, value, salt);
+    }
+    
+    /*
+    ## Blind auction for the desired hash
+
+    Bids are sent by sending a message to the main contract with a hash and an amount. The hash 
+    contains information about the bid, including the bidded hash, the bid amount, and a random 
+    salt. Bids are not tied to any one auction until they are revealed. The value of the bid 
+    itself can be maskeraded by changing the required period or sending more than what you are
```
Smithgift: s/maskerade/masquerade/

HashRegistrar.sol
```diff
@@ -0,0 +1,313 @@
...
+    itself can be maskeraded by changing the required period or sending more than what you are 
+    bidding for. This is followed by a 24h reveal period. Bids revealed after this period will 
+    be burned and the ether unrecoverable. Since this is an auction, it is expected that most 
+    public hashes, like known domains and common dictionary words, will have multiple bidders pushing the price up. 
+    */ 
+    function newBid(bytes32 sealedBid) {
+        if (address(sealedBids[sealedBid]) > 0 ) throw;
+        // creates a new hash contract with the owner
+        address newBid = new Deed();
+        sealedBids[sealedBid] = Deed(newBid);
```
Smithgift: I think `new Deed()` would return a Deed, so you wouldn't need to recast it.

HashRegistrar.sol
chriseth: Indeed, there are two ways to exploit this here: Either the owner can throw in the fallback function or the registrar can perform a call stack depth attack. The only known way to prevent both is the withdraw pattern.

Arachnid: The owner throwing doesn't do them any good; it just ensures they don't get their ether. Good point on the stack depth attack.

HashRegistrar.sol
```diff
@@ -47,34 +47,52 @@ contract Deed {
     address constant burn = 0xdead;
     uint public creationDate;
     address public owner;
-    
+    event OwnerChanged(address newOwner);
```
Arachnid: I'm concerned these events would be difficult to listen for, since they're on each deed; can we sensibly restrict the events to the registrar contract?

HashRegistrar.sol
```diff
@@ -47,34 +47,52 @@ contract Deed {
...
+    event deedClosed();
```
Arachnid: Either this should be `DeedClosed` or the other one should be `ownerChanged`.

HashRegistrar.sol
```diff
@@ -47,34 +47,52 @@ contract Deed {
...
-        owner.send((refundRatio * this.balance)/1000);
-        selfdestruct(burn);
+    function closeDeed(uint refundRatio) onlyRegistrar onlyActive {
+        active = false;            
+        burn.send(((1000 - refundRatio) * this.balance)/1000);
+        deedClosed();
+        destroyDeed();
     }    
+
+    function destroyDeed() {
```
Arachnid: I wonder if changing to a pure withdrawal model wouldn't make this simpler: have `setBalance` only update a balance variable, and have a `withdraw` method, only callable by the owner, that sends them any extra and self-destructs iff the recorded balance is zero (on the assumption that deposits are always nonzero).

That would restrict sends to just that method, and remove the need to reason about who's calling things like setBalance and whether they have incentive to make life difficult; on the other hand, it would require callers to interact with their deeds directly on a day-to-day basis.

What do you think?

HashRegistrar.sol
```diff
@@ -91,7 +109,12 @@ contract Registrar {
     uint public averagePeriod;
     uint public lastSinceNewRegistry;
     uint public averagePrice;
-    
+
+    event AuctionStarted(bytes32 hash);
```
Arachnid: It might be worth making `hash` and `owner` indexed in these events.

HashRegistrar.sol
```diff
@@ -91,7 +109,12 @@ contract Registrar {
...
+    event AuctionEnded(bytes32 hash, address owner, uint value, uint averagePrice, uint averagePeriod);
```
Arachnid: Is it necessary to log averagePrice and averagePeriod here?

HashRegistrar.sol
```diff
@@ -200,13 +222,15 @@ contract Registrar {
...
             // cancel the other bid, burn 0.1%
-            Deed previousWinner = Deed(h.deed);
+            Deed previousWinner = h.deed;
             previousWinner.closeDeed(999);
             
             // set new winner
             h.value = h.highestBid;
             h.highestBid = _value;
             h.deed = sealedBids[seal];
+            bid.setOwner(_owner);
```
Arachnid: setOwner should still be up top, otherwise the other conditions that close it won't return any ether to them.

HashRegistrar.sol
```diff
@@ -0,0 +1,340 @@
...
+        Deed bid = sealedBids[seal];
+        if (address(bid) == 0 ) throw;
+        sealedBids[seal] = Deed(0);
+        
+        entry h = entries[_hash];
+        
+        if (bid.creationDate() > h.auctionExpiration - revealPeriod
+            || now > h.auctionExpiration ) {
+            // bid is invalid, refund 99.9%
+            bid.closeDeed(1);
```
holiman: This burns 99.9%, so refunds 0.1%, contrary to what the comment says

HashRegistrar.sol
```diff
@@ -0,0 +1,340 @@
...
+        } else {
+            // bid doesn't affect auction
+            bid.closeDeed(999);
+        }
+    }
+    
+    function cancelBid(bytes32 seal) {
+        Deed bid = sealedBids[seal];
+        if (address(bid) == 0 || now < bid.creationDate() + auctionLength * 2 || bid.owner() > 0) throw; 
+        bid.closeDeed(0);
```
holiman: This burns all ether. Is that the intention? And why can't bids be canceled before 2*auctionlength?

Arachnid: Yes; this permits someone other than the bidder to delete bids for long-expired auctions. The delay is so that the legitimate bidder has an opportunity to reveal their bid and recover their ether.

HashRegistrar.sol
```diff
@@ -0,0 +1,340 @@
...
+        burn.send(((1000 - refundRatio) * this.balance)/1000);
+        deedClosed();
+        destroyDeed();
+    }    
+
+    function destroyDeed() {
+        if (active) throw;
+        if(owner.send(this.balance)) 
+            selfdestruct(burn);
+    }
```
holiman: Wouldn't burn.send then selfdestruct(owner) be more robust? Otherwise, owner can prevent destruction.

HashRegistrar.sol
holiman: Ah I see. But there's no incentive (in fact, there's a cost) in cleaning up that way. Why not credit cleaner instead of burning it all?

holiman: Or better,  send bulk to owner, percent to cleaner, some to burn (maybe)

HashRegistrar.sol
Arachnid: The problem with that is that selfdestructs don't run fallback code, so people with wallets won't get an event for the funds being sent. I  think that's worth some trouble to avoid breaking.

HashRegistrar.sol
```diff
@@ -0,0 +1,340 @@
...
+    /* 
+    The Deed is a contract intended simply to hold ether
+    It can be controlled only by the registrar and can only send ether back to the owner.
+    */
+    Registrar public registrar;
+    address constant burn = 0xdead;
+    uint public creationDate;
+    address public owner;
+    event OwnerChanged(address newOwner);
+    event deedClosed();
```
larspensjo: Naming convention is capital 'D' in `deedClosed`?

HashRegistrar.sol
Smithgift: "Cleaner" will most likely be "Guy whose transaction has the highest gasPrice and therefore their cleaning transaction is executed first in the relevant block." (Alternately, the miner) The bid still gets cleaned, but the money would go to a strange place.

jcnelson:
We did an independent code review last night, and came up with a couple of higher-level points.  We were a bit paranoid, so feel free to ignore if these aren't actual concerns.

* It's easy to accidentally misuse this contract in a way that removes the bid blindness.  The fact that shaBid() is public means that an unknowing user might accidentally disclose the hash and salt by using it to generate a sealed bid.

* Bids probably won't happen until the name hash is revealed.  This is because the hash can have many different seals.  This tells me that there's no incentive to reveal a newly-registered name hash until the very end of its auction period, but there's every reason to keep it secret until the last minute to ensure a lower deposit.

* Bidding on multiple name hashes at once is not a particularly good blinding strategy, since (1) no one would place bids until the name hash was revealed, and (2) the space of desirable names (and their hashes) is easily pre-computed in most problem domains (i.e. DNS names, human names, dictionary words, brand names, etc.).

* I don't think using a deposit model does enough to stop squatting.  The proposition looks pretty low-risk from the squatter's perspective:  buy a bunch of name hashes, and either get all the ETH back with `releaseDeed()` or get a profit by selling the name's private key before `h.firstRegistered + renewalPeriod/2`.  On the other hand, if you had to burn the winning bid, you'd be incentivized to either use the name somehow (to recoup your investment), or be strategic on what names you'll buy if you want to re-sell them.

Arachnid:
> We did an independent code review last night, and came up with a couple of higher-level points. 

Thanks for taking the time to give it a thorough review!

> It's easy to accidentally misuse this contract in a way that removes the bid blindness. The fact that shaBid() is public means that an unknowing user might accidentally disclose the hash and salt by using it to generate a sealed bid.

This would be difficult for users to do; by default `constant` methods are executed on the local node; if you manually execute it as a transaction instead, you can't retrieve the return value!

> Bids probably won't happen until the name hash is revealed. This is because the hash can have many different seals. This tells me that there's no incentive to reveal a newly-registered name hash until the very end of its auction period, but there's every reason to keep it secret until the last minute to ensure a lower deposit.

I'm not sure I follow; the name hash for a particular bid is revealed in the auction reveal phase, but at that point it's too late to place new bids. You would not normally unblind your bid while the auction is still ongoing.

> I don't think using a deposit model does enough to stop squatting.

Squatting's definitely problematic. This system won't prevent squatting on names that gain in value faster than the market as a whole, but it does at least eliminate the incentive to squat on vast numbers of names that only appreciate in value as fast as the system as a whole does.

I think the existing DNS system demonstrates that 'burning' registration fees isn't itself enough to discourage squatting, so I'm skeptical that would work any better than this.

HashRegistrar.sol
holiman: I'm not sure I understand what you mean, could you elaborate? 

What I mean is that if we _do_ want clean-up to be performed (and we do), then there must be a reward for cleaning up, not a cost. And since we're burning money, why not use some of it for something useful instead? We'd still need to burn the same amount as if the `owner` had revealed it, to not create a game:able situation. I don't see a problem with there being 'race to clean'.

HashRegistrar.sol
holiman: Ok, well not letting it selfdestruct if `send` fails accomplishes nothing, imho. The `active` has been set to false, so `closeDeed` can't be called again anyway. 

A better solution may be sth like this:  


       if(!owner.send(this.balance)) selfdestruct(owner)
        selfdestruct(burn);

jcnelson:
> This would be difficult for users to do; by default constant methods are executed on the local node; if you manually execute it as a transaction instead, you can't retrieve the return value!

Ah, thank you, that clarifies things :)  Updated my comment.

> I'm not sure I follow; the name hash for a particular bid is revealed in the auction reveal phase, but at that point it's too late to place new bids. You would not normally unblind your bid while the auction is still ongoing.

I think there are couple things I'm still not sure on:

* In `startAuction()`, I guess I'm not sure what `_hash` is supposed to be here.  Is it the hash of the name I'm bidding on, or is it a seal?  I assumed the latter, since the former would reveal the name hash right at that point.  But since you're using blind bidding, I'm guessing that's not really a concern?

* Is the point of sealing to blind the `value`, the `hash`, or both?  It looks like it only blinds the `value`, given the above.

> Squatting's definitely problematic. This system won't prevent squatting on names that gain in value faster than the market as a whole, but it does at least eliminate the incentive to squat on vast numbers of names that only appreciate in value as fast as the system as a whole does.

I like the idea of having prices adjust based on demand :)  However, it's not clear to me that this works better than ICANN if you're depositing instead of buying.  All the analysis I can find on second-price sealed bid auctions assume that the winner *pays* the money.

More generally, is blind bidding the right approach here?  Why is price discovery *not* important?

Arachnid:
> In startAuction(), I guess I'm not sure what _hash is supposed to be here. Is it the hash of the name I'm bidding on, or is it a seal? I assumed the latter, since the former would reveal the name hash right at that point. But since you're using blind bidding, I'm guessing that's not really a concern?

`_hash` is the hash of the name you're opening the auction for. If it's a name with an obvious preimage, then others will know that _someone_ is bidding on that name. Being able to open multiple auctions cost-free is supposed to reduce the risk of people just speculatively bidding on everything that gets opened.

> Is the point of sealing to blind the value, the hash, or both? It looks like it only blinds the value, given the above.

It blinds the value, but it also conceals which open auction you're bidding on until that auction's bidding phase has concluded.

> More generally, is blind bidding the right approach here? Why is price discovery not important?

Good question, and probably one for @alexvandesande to answer; I would be equally onboard with unblinded bids and a wait-for-quiet auction model; I think both have things to recommend them.

csuwildcat:
In terms of pricing names, one way to set a floor for names which already are regarded as high value in today's domain system is to leverage crawler data to establish a price that reflects the current state of demand traffic for a name/word using something like Google Keyword Planner. Keyword Planner givves you the average monthly search volume for a provided term, which is a far better indicator of popularity than some bespoke auction participant demand level within still nascent systems like the ones we're all building.

If you can find a way to relatively score the incoming name against the demand being expressed in the market today, you have a much better basis for establishing a price floor.

That's my 2 cents.

jcnelson:
> _hash is the hash of the name you're opening the auction for. If it's a name with an obvious preimage, then others will know that someone is bidding on that name. Being able to open multiple auctions cost-free is supposed to reduce the risk of people just speculatively bidding on everything that gets opened.

Yup, and there can't be multiple concurrent auctions on the same hash.

However, besides the gas fees, what stops an attacker from opening an auction for every word in the dictionary, bidding a minimal amount on each one, and then waiting for the auction to finalize?  Wouldn't he basically acquire a bunch of names for very cheap then?  There's little downside risk of not being able to re-sell them, since he can just get his deposit back.  Moreover, very few potential users of this smart contract know it exists, let alone how to interact with it, so he has a high chance of successfully grabbing a lot of names.

> It blinds the value, but it also conceals which open auction you're bidding on until that auction's bidding phase has concluded.

A few things here:

* Can you guarantee that (1) addresses used to generate the hash are never-before-used, and (2) the salt is never-before-used?  If either one of these ever gets reused across multiple `revealBid()` calls, then calculating the seal's preimage reduces to search the value space (which isn't that big).  This lets me unblind values and hashes before `revealBid()`.

* Related, a bidding client should be careful to use a different address to generate the seal than the one used to send the transaction that calls `newBid()`.  Otherwise, I can get the address before the `revealBid()` call, which reduces the attacker's search space.

* Will calculating time-based bid frequency distribution give me any insights into the likely distribution of names currently being bid on?  I strongly suspect so.

HashRegistrar.sol
Smithgift: Race-to-clean isn't a _big_ issue, but the money is going to, most likely, the wealthiest cleaner. I would prefer if there was another way, but I don't object to paying cleaners if this is the only practical option.

Smithgift:
Wait-for-quiet has form of a "Bobcat" attack, where the attacker bids on anything that's been bidded on on the basis that _someone_ wants it (or just to be a jerk). Countertrolls would counterattack by bidding on worthless hashes to trick Bobcats into buying them. Then Bobcats stop bidding on random hashes, and countertrolls waste their money. Then things get complicated. There's some equilibrium, most likely, but it may not be pleasant for the users.

Here, no one knows that a given hash is valuable until it's over. It may be worse in that someone sneaks off with wedding-shoes.eth for cheap, but the average user may suffer less.

I'm hoping someone will construct a public rainbow table of every dictionary word, name, etc. Then it's easy enough to examine the auction list and see if anyone's buying one. There's going to be a gold rush, but it could hopefully be fairer one.

There was talk in an earlier thread of banning the hashes of small names for early days (and discussion of how that could even be done.)

Arachnid:
> However, besides the gas fees, what stops an attacker from opening an auction for every word in the dictionary, bidding a minimal amount on each one, and then waiting for the auction to finalize?

A minimum price is set, currently at 1/100th of the average, though we'd like a better mechanism for that. Besides that, the attacker/squatter's deposits are tied up for a minimum of 6 months before he can release the domain and recover the deposit.

> Moreover, very few potential users of this smart contract know it exists, let alone how to interact with it, so he has a high chance of successfully grabbing a lot of names.

I expect that by the time this is ready to launch on mainnet, it will be significantly higher profile.

> Can you guarantee that _owner and _salt are never-before-used? If either one of these ever gets reused across multiple revealBid() calls, then calculating the seal's preimage can reduce to searching the value space and spaces of known addresses and owners (which isn't that big).

It's up to the bidder to generate a sufficiently random salt to prevent this, but I don't think that's particularly onerous.

> Related, a bidding client should be careful to use a different _owner address than the one used to send the transaction that calls newBid(). Otherwise, the attacker would get the address before the revealBid() call, which reduces his search space.

As long as the salt is chosen using high quality randomness, I do not think this is a concern.

> There was talk in an earlier thread of banning the hashes of small names for early days (and discussion of how that could even be done.)

Excellent point. Doing this would require the winner to reveal the preimage of the hash they're 'buying'.

HashRegistrar.sol
alexvandesande: But this means there's a higher risk of the ether being lost. With the current method, even with `active=false` you can still change the owner and try to send it somewhere else.

HashRegistrar.sol
alexvandesande: I've considered this, but right now the whole architecture is built so that the deed can *only send money to the owner*. With this, I'd have to open up some sort of exception that makes it possible for the ether to be sent elsewhere. I'm afraid this opens up a breach that could be exploited somehow..

alexvandesande: Mmm. Come to think about it, not necessarily: the registrar could change the owner to the msg.sender and then close the deed..

HashRegistrar.sol
```diff
@@ -148,17 +164,22 @@ contract Registrar {
...
         
         if (newAuction.status == Mode.Owned) {
             Deed deedContract = Deed(newAuction.deed);
             deedContract.closeDeed(999);
         }
         
-        newAuction.auctionExpiration = now + auctionLength;
+        // for the first six months of the registry, make longer auctions
+        // TODO: change 'hours' into 'weeks' for real contract
+        uint slowStart = 1 + (registryCreated + 20 hours - now) / 4 hours;
```
Smithgift: You may be able to use a pattern where slowStart is read from a state variable. Said variable would be calculated in a separate function, pinged by kindly users who want their auctions to end sooner.

Having said that, the optimization may be so small that it's not worth the additional complexity.

Arachnid: Would a simple threshold scheme not work here? Eg, auctions last 2 weeks for the first 6 months, then revert to 1 week (overnight)?

HashRegistrar.sol
```diff
@@ -100,28 +100,29 @@ contract Registrar {
...
     uint public lastSinceNewRegistry;
-    uint public averagePrice;
+    uint public registryCreated;
 
-    event AuctionStarted(bytes32 hash);
-    event AuctionEnded(bytes32 hash, address owner, uint value, uint averagePrice, uint averagePeriod);
-    event DeedRenewed(bytes32 hash, uint value, uint renewalDate);
-    event DeedReleased(bytes32 hash, uint value);
+    event AuctionStarted(bytes32 hash, uint auctionExpiryDate);
+    event BidRevealed(bytes32 hash, address owner, uint value, uint8 status);
```
Arachnid: Shouldn't status be an enum?

Arachnid: `hash` and `owner` should probably be indexed, too.

HashRegistrar.sol
```diff
@@ -202,44 +223,49 @@ contract Registrar {
...
+            // cancel the other bid, refund 99.9%
             Deed previousWinner = h.deed;
             previousWinner.closeDeed(999);
             
             // set new winner
             h.value = h.highestBid;
             h.highestBid = _value;
-            h.deed = sealedBids[seal];
+            h.deed = bid;
             bid.setOwner(_owner);
```
Arachnid: `setOwner` needs to be outside the else clause

alexvandesande:
@jcnelson thanks for the insightful comments!

> I don't think using a deposit model does enough to stop squatting. The proposition looks pretty low-risk from the squatter's perspective: buy a bunch of name hashes, and either get all the ETH back with releaseDeed() or get a profit by selling the name's private key before h.firstRegistered + renewalPeriod/2.

Of course, this has never been tested before and we are not sure it is actually effective on preventing squatters, but let's play a scenario here. Suppose a squatter buys up a 10,000 names for the minimum bid (let's say it's 0.01 ether). This means he will have to lock up 100 ether for at least 6 months. What happens to the average price of domains in one year?

* It goes down 50%: Mr. Squatter can renew all his names AND get 25 ether back! Sounds like a great deal, until they realize that had he bought the same amount of names for a minimum bid now he would only have spent 50 ether. 

* It goes up 50%: Mr. Squatter will need to add another 50 ether to keep them. Notice that the if he would buy the same amount of names for the minimum bid now he would have paid the same amount. 

* It stays about the same. Mr. Squatter can keep the names for free, but he has been holding all that ether for names no one wants: Had he staked it (or lent it or anything more productive) he would have a better profit. If he is not doing anything, he has a good incentive to release them so they can be auctioned off again.

Of course this assumes names are an near infinite commodity and that there always will be new names to buy, and this assumes a healthy ecosystem where there are better things to do with ether that is just being held.

> There was talk in an earlier thread of banning the hashes of small names for early days (and discussion of how that could even be done.)

It's not that easy to do currently and length is not a great indicator of popularity. "weddings.eth" is probably a more valuable name than "xt2g1.eth". There are two ways I plan on helping the issue of name popularity:

* Pre-populate the default name registry app with a lot of company names and dictionary words, so the interface would show clearly that the auction on "apple.eth" or "bank.eth" has started, thus incentivising more people to bid on it.

* Make longer auctions on the beginning. While the goal would be a week long auction, for the first six months of the system running, probably we should have longer 6-4 week auctions, so that the most obvious names will have a larger audience of people bidding on them.

The goal is to make so that a made up name like `ethereum-text-adventure-game.eth` could be acquired for almost nothing, while popular names will have a high bid - so that the owners of the most popular names will have to pay larger amounts to keep their names.

csuwildcat:
I will note one very important thing for the initiative underway between some of the blockchain identity folks here and the organization I represent: The UX of a auction-based, time-delayed registration scheme is unacceptable for most consumer ID acquisition use-cases. I know from my perspective as the lead of our decentralized identity group, I would never allow a user to be exposed to a system/UX like that, which would effectively take Ethereum-rooted names out of the running for use - not something I want to do.

I strongly encourage folks here to explore options for instant valuation floor discovery via utilization of empirical market demand data provided by outlets like Alexa, which can be used to set accurate, effective price floors for high value names, all without building a Rube Goldberg Machine of unnecessary complexity, fraught with such fundamental issues.

HashRegistrar.sol
```diff
@@ -0,0 +1,374 @@
...
+    function setBalance(uint newValue) onlyRegistrar onlyActive {
+        // Check if it has enough balance to set the value
+        if (this.balance < newValue) throw;
+        // Send the difference to the owner
+        if (!owner.send(this.balance - newValue)) throw;
+    }
+
+    function closeDeed(uint refundRatio) onlyRegistrar onlyActive {
+        active = false;            
+        burn.send(((1000 - refundRatio) * this.balance)/1000);
```
colm: Do we need to worry about stack depth attack here? I can see a situation where a registrar and owner wants `closeDeed` to fail so that they can call `destroyDeed` to get `this.balance` thus they can avoid the burning.

holiman: I think you're right, good catch!

Arachnid: But both the send to burn and the send to owner are at the same level; they can't cause one to fail without the other.

Arachnid:
> I will note one very important thing for the initiative underway between some of the blockchain identity folks here and the organization I represent: The UX of a auction-based, time-delayed registration scheme is unacceptable for most consumer ID acquisition use-cases. I know from my perspective as the lead of our decentralized identity group, I would never allow a user to be exposed to a system/UX like that, which would effectively take Ethereum-rooted names out of the running for use - not something I want to do.

I fully expect that most users will not be registering 'foo.eth' domains themselves, just like end users don't typically register DNS domains themselves; instead, I expect secondary registrars - say, 'joe.wallet.eth' or 'joe.me.eth' - will fill that role with simple cost-free first-in-first-served registration.

> I strongly encourage folks here to explore options for instant valuation floor discovery via utilization of empirical market demand data provided by outlets like Alexa, which can be used to set accurate, effective price floors for high value names, all without building a Rube Goldberg Machine of unnecessary complexity, fraught with such fundamental issues.

If you think this is Rube Goldbergian, you haven't seen what a system that relies on an outside oracle for price discovery would look like!

If you have an alternate proposal, though, we'd very much welcome it - the governance of .eth is far from set in stone, and PRs are welcome.

csuwildcat:
Take a securely attained, periodic snapshot of traffic level (the primary name demand indicator) for the top X number of domains and hash-anchor the data to the blockchain for verifiable reference. When a name registration attempt is made, check for presence via an oracle that does lookups against the flat mapping and responds with data (if present) and the verification proof. If data is returned showing the name is of higher-than-minimum value, compute a price based on the demand expression of the existing, worldwide market.

Issue #1

I'm curious: for the period of time an auction mechanisms for a nascent system with little outside interest from the public is registering names, how do you plan on overcoming the 'If a Name Falls in the Woods' problem?

Example: Assume I announce I want to register 'google' and very few people (certainly not Google) are around who care, given relatively few people are using the system at the time. I would expect the few people in the system during its early stages of obscurity will bid that up a little, but nothing like if it were Google.com selling in the current ICANN market. So I acquire Google.com for relative pennies because my 'public' announcement has the effect of yelling "Google" in the woods - and this is prevented how?

Issue #2

Say, just hypothetically now, I'm a large consumer technology company implementing a universal, open source, cross-blockchain system of decentralized ID registration. In this system I have other blockchains that can assess value floors and issue their TLDs in a few hours. One of my product requirements is root domain registration, not just a subdomain as you highlighted, must take no more than a few hours. How do I include your system as an option? Do I put a big red warning sign in the UI saying "Users who choose TLDs rooted in this system can expect to wait X days for notification of the success or failure of their registration"?

Arachnid:
> Take a securely attained, periodic snapshot of traffic level (the primary name demand indicator) for the top X number of domains and hash-anchor the data to the blockchain for verifiable reference.

I think there's a couple of problems with that; one is that such metrics are proprietary and not freely available, as well as being of questionable reliability; the other is that it assumes that the same names will be popular on Ethereum as on the DNS. I don't think a measure like that which relies on using external data as a barometer is going to be popular as an internal governance measure.

> I'm curious: for the period of time an auction mechanisms for a nascent system with little outside interest from the public is registering names, how do you plan on overcoming the 'If a Name Falls in the Woods' problem?

It's a problem alright, but all we can plausibly do is ensure the system is as widely known as possible before name registration on mainnet becomes possible. If you have a suggestion on how to remediate it, we're all ears.

> Say, just hypothetically now, I'm a large consumer technology company implementing a universal, open source, cross-blockchain system of decentralized ID registration. In this system I have other blockchains that can assess value floors and issue their TLDs in a few hours. One of my product requirements is root domain registration, not just a subdomain as you highlighted, must take no more than a few hours. How do I include your system as an option? Do I put a big red warning sign in the UI saying "Users who choose TLDs rooted in this system can expect to wait X days for notification of the success or failure of their registration"?

You could offer instantaneous resolution for 'anything.mysite.eth' by setting up your own resolver; if your users want to register first-class names, they'll have to play by the same rules as everyone else.

Again, this is still very much a work-in-progress; you're very welcome to send in PRs with your own governance proposals, or write up a spec and encourage others to implement it.

Smithgift:
In, re: the $TRADEMARK.eth, I suspect anyone who buys google.eth will find later themselves in contact with Google's legal department if Google ever decides they want it. That doesn't mean someone won't buy it anyway.

Unfortunately, there's no practical way to reserve every trademark (let alone everything that _could_ infringe on a trademark). A merkle tree could theoretically hold _a_ list, but I cannot conceive of how to construct said list. See also: the potential legal issues, which I am not qualified to even speculate on.

larspensjo:
You can be a completely anonymous owner? Which means there is no way to
enforce outside world trademarks.

Smithgift:
I believe that Google (or Nike/Coca Cola/etc) would try their hardest to find a way. For example, transaction analysis to try to find who paid for google.eth. Alternately, attempt to get the court to enforce the trademark by mandating a fork with irregular state transition. (Insert community-wide crisis here.) 

A simpler option may be for Google to convince the network to create a *.google "TLD" for them, End users will have to be trained not to trust something with google in the name that isn't a *.google.

csuwildcat:
@Arachnid I can basically guarantee reliable, accurate demand/traffic data will be freely available for such an implementation - we have access to this data. As far as "outside" data being frowned upon by the community: is the Ethereum community actually taking the position that empirical, objective data that doesn't originate from activity on its chain is not to be used or relied upon? That seems like a strange and unrealistic view of how to implement a wide array of functionality that fundamentally and beneficially reliant on off-chain data.

In response to your answer for Issue 1: I can almost assure folks that much of the current ICANN domain space will want to secure the same equivalent in any new naming system. Additionally, I am highly skeptical (as any technical PM should be) that an emerging system will have the network effect and social involvement to support a scheme that relies on the flywheel running at speed on Day 1.

What I heard as a response to Issue 2 is a significant problem for us. We would either exclude or minimize the presence of a registration flow to users for _any_ systems that require subdomain issuance, for a number of reasons:

1. It requires issuance by an entity
2. The root name upon which subs are issued is decided by the issuing entity
3. It effectively eliminates the ability for non-technical users to secure root names
4. It doesn't fit with the UX model we would build our implementation around

We would still happily index names that end up being rooted in the Ethereum chain, but the loss of a registration flow exposed prominently in our products/services would be unfortunate - I would like to avoid that at all costs.

If these things don't matter to this community, that's fine - it's not my call. I am just trying to provide product/customer feedback that, hopefully, helps address the needs of developers, groups, or organizations who would like to utilize the system.

Arachnid:
> I can basically guarantee reliable, accurate demand/traffic data will be freely available for such an implementation - we have access to this data.

Who is we, and how can people verify that the data is accurate?

> As far as "outside" data being frowned upon by the community: is the Ethereum community actually taking the position that empirical, objective data that doesn't originate from activity on its chain is not to be used or relied upon?

Not at all. I've got two concerns: first, that the data doesn't necessarily apply cleanly to the situation at hand; I think there's a strong positive correlation, but don't expect that values of DNS domains and ENS domains will match in all cases; second, that insofar as the data requires special access of some kind to gather, it constitutes a central point of failure and/or control for a nominally distributed system.

> We would either exclude or minimize the presence of a registration flow to users for any systems that require subdomain issuance, for a number of reasons:

> It requires issuance by an entity

Can you clarify what you mean by that? A subdomain registrar can be set up in exactly the same mechanism as a domain registrar.

> It effectively eliminates the ability for non-technical users to secure root names

I fully expect services to spring up to make registering names easier than the base system entails; even initially we're likely to have a DApp that streamlines the process, and it sounds like you're in the business of building something not dissimilar. The only remaining factor that may prove a barrier to users is the week-long registration time.

This is straying a bit offtopic for this (now submitted!) PR; would you be open to opening an issue for further discussion?

csuwildcat:
@Arachnid perhaps we should just chat directly about the initiative underway at Microsoft (and include Christian/Rouven).  If you feel you would like to have this implementation included for consideration after hearing about what's in the works, I can open GH Issues for the blockers this type of mechanism represents.

HashRegistrar.sol
colm: :+1: 
PR https://github.com/alexvandesande/ens/pull/1 :D

Arachnid:
Sure; please email me at nick@ethereum.org!

HashRegistrar.sol
alexvandesande: Can you describe better how this could fail? `burn.send` sends ether to `0xdead00000..` address so I can't see how that would fail and even if it does it doesn't actually care about the return values.

Is it possible for a stack depth attack to execute the `active = false` but stop before it does the `burn.send`? If that's the case, then I could simply switch the two, but then the opposite problem would arrive: someone could use the same attack to burn money repeatedly.

alexvandesande:
@csuwildcat  I feel your current description, relying on external data for domain values, is much more rube goldbergian. It also ties these values to domains, and we don't actually know they are tied to these. Instead, just using a simple auction mechanism allows the names to be tied to real demand.

Actually, it doesn't exclude what you are saying: someone could built a smart bot that uses any sort of proprietary data and private algorithm to try to assess the value of each name, and bid on them accordingly. The better they are than the rest of the market, the more they will be rewarded by owning valuable names.

The only real disadvantage here is the fact that names take a week to be registered. I simply don't see any way around it without changing drastically the auction model. But if the project works as intended, this could encourage a domain business to preemptively buy names they believe would be valuable and hold them for six months to try to sell them with instant registration. In theory, the difference with the proposed system is that it would encourage these resellers to release names constantly back to the market, in order to free up stake ether, instead of holding all forever.

> how do you plan on overcoming the 'If a Name Falls in the Woods' problem?

First of all this contract is just part of the solution. Another would be building an easy ethereum app to use them, and that would probably include a sizeable file with hashes of popular trademarks, domains and dictionary words, so that these auctions would be broadcasted to everyone using the system.

Of course I don't expect the world to be using it right away, but that's where averages play nicely: if you buy names when the system isn't used and names for for cheap, but then one year later, when there are tons of people using it, the average price goes way up, the amount of ether required for you to hold the names would also increase, meaning there's a lot more incentive for you to re-release the names back to the wild.

HashRegistrar.sol
colm: http://ethereum.stackexchange.com/questions/6150/how-does-the-stack-depth-attack-make-a-send-silently-fail

So basically the max stack depth is 1024. So before the attacker calls the function she creates a stack of 1023 by repeatedly calling proxy contracts. Then the function is called (increassing the stack to 1024) and `burn.send` adds one to the stack which is illegal so this send fails silently. But does not throw. The contract continues is execution. 

There has been some discussion in the solidity gitter about changing this behavior. But for now all sends should be checked in like `if !(send.burn()) throw;`

PR #2, opened 2016-08-27
add a default function to Deed contract

pirapira:
Since `Registrar` contract contains an expression `newBid.send(msg.value)` for a `Deed newBid`, the `Deed` contract needs a fallback function.  This pull-request adds a fallback function that just returns.


Arachnid:
The default functionality is to just receive ether. This will change in the next version of Solidity, but only if you specify a pragma. Nevertheless, I'm going to merge this just because explicit is better than implicit.

PR #3, opened 2016-08-27
Possibly more correct `renewalDate` computation

pirapira:
The `renewalDate` computation contained an expression `2 * now - h.registrationDate`.  Supposing `now` is year 3000 (the code should still work in the 30th century), this expression evaluates somewhere around year 4030.

This commit tries to figure out the original intent.

### DIFF
--- a/HashRegistrar.sol
+++ b/HashRegistrar.sol
@@ -344,7 +344,8 @@ contract Registrar {
         h.value = updatedPrice;
         h.lastRenewed = now;
         // Twice the current age, as long as it's betwen some max and min parameters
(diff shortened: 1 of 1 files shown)


pirapira:
Pinging @alexvandesande because I landed here from his reddit post.

Arachnid:
Good spotting. This eliminates the `max`, which was intended to make sure a renewal is at least one ordinary renewal period, though.

pirapira:
When `currentAge` is less than 12 hours, the `max` is still effective.

Arachnid:
Oops, I misread the diff. Thanks!

pirapira:
Ah, now that's clear.  Sorry for the noise.  `2 * date - date` is a `date`.

PR #5, opened 2016-08-29
Preventing subtraction underflow in slowStart

pirapira:
Before this pull-request, the computation of `slowStart` contained a subtraction that can underflow:
```
registryCreated + 20 weeks - now
```
when the current timestamp is more than 20 weeks after the registry creation.  In unlucky cases the `newAuction.registrationDate` would become a point in the past.  This pull request avoids the subtraction when the underflow would occur.

I hope this time it's not an arithmetic misconception (like the one I made in #3).

### DIFF
--- a/HashRegistrar.sol
+++ b/HashRegistrar.sol
@@ -177,7 +177,9 @@ contract Registrar {
         }
         
         // for the first six months of the registry, make longer auctions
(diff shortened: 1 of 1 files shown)


alexvandesande:
Thanks for that. Being an `uint` I thought that it would automatically convert any negative number into `0`.

PR #8, opened 2016-09-05
Moved inline check for owner into a modifier.

zigguratt:
For consistency within this repo, I moved the inline check for `subnode` ownership in `FIFSRegistrar.sol` from within the `register` function into an `only_owner` modifier.


Arachnid:
Thanks!

PR #9, opened 2016-09-13
Added temporary simplified registrar

alexvandesande:
A time and character temporary registrar


HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+    function closeDeed(uint refundRatio) onlyRegistrar onlyActive {
+        active = false;            
+        if (! burn.send(((1000 - refundRatio) * this.balance)/1000)) throw;
+        DeedClosed();
+        destroyDeed();
+    }    
+
+    function destroyDeed() {
+        if (active) throw;
+        if(owner.send(this.balance))
```
Arachnid: This should probably throw if it fails, for consistency.

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+    ENS public ens;
+    bytes32 public rootNode;
+
+    mapping (bytes32 => entry) public entries;
+    mapping (bytes32 => Deed) public sealedBids;
+    
+    enum Mode { Open, Auction, Owned, Forbidden }
+    uint32 constant auctionLength = 7 days;
+    uint32 constant revealPeriod = 24 hours;
+    uint constant minPrice = 0.01 ether;
```
Arachnid: As an alternative to a hard deadline, we could implement a difficulty-bomb style exponential increase of the minimum bid. What do you think?

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+    uint public registryCreated;
+
+    event AuctionStarted(bytes32 hash, uint auctionExpiryDate);
```
Arachnid: Parameter names here should probably distinguish 'masked hash' from 'name hash'.

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+    event NewBid(bytes32 hash, uint deposit);
+    event BidRevealed(bytes32 hash, address owner, uint value, uint8 status);
```
Arachnid: `hash` and `owner` should probably be indexed.

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+
+    struct entry {
+        Mode status;
+        Deed deed;
+        uint registrationDate;
+        uint value;
+        uint highestBid;
+    }
+    
+    modifier noEther {
```
Arachnid: If you update this contract to latest solidity, this is no longer required; instead, you have to specify 'payable' on functions that should accept ether.

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    function  min(uint a, uint b) internal constant returns (uint min) {
+        if (a < b)
+            return a;
+        else
+            return b;
+    }
+
+    /*
```
Arachnid: I would actually do away with the slice entirely, and just count characters on the actual string.

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+    */    
+    function startAuction(bytes32 _hash) noEther {
+        entry newAuction = entries[_hash];
+        if ((newAuction.status == Mode.Auction && now < newAuction.registrationDate)
+            || newAuction.status == Mode.Owned 
+            || newAuction.status == Mode.Forbidden
+            || now > registryCreated + 4 years)
+            throw;
+        
+        // for the first five months of the registry, make longer auctions
```
Arachnid: Weren't we going to change to an extended period on launch, then transition straight to 1 week auctions?

HashRegistrarSimplified.sol
Arachnid: Is it possible to make the last argument of type Mode, or another enum?

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+        || now > registryCreated + 8 years) throw;
+        
+        h.status = Mode.Open;
+        ens.setSubnodeOwner(rootNode, _hash, 0);
+        deedContract.closeDeed(1000);
+        HashReleased(_hash, h.value);
+    }  
+
+    /*
+    Names on the simplified registrar can't be six letters or more. We are purposefully
```
Arachnid: Can't be six letters or _less_.

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,367 @@
...
+    handicapping it's usefulness as a way to force it into being restructured in a few years
```
Arachnid: s/it's/its/

HashRegistrarSimplified.sol
alexvandesande: Then I didn't understand how it works because the len() function takes a slice, not a string

Arachnid: Right. But the code can be changed to iterate directly over the string, without needing to create a slice first. The slice abstraction is there to make repeated operations over strings efficient without having to copy the data each time.

HashRegistrarSimplified.sol
alexvandesande: What would be the advantage? In a worst case scenario, if in 4 years we aren't able to deploy a  new registrar code, all it would do is prevent new names from being registered, meaning it could still be usable - and more importantly, they can still be converted into the new thing later..

HashRegistrarSimplified.sol
alexvandesande: But these are the same. It's an auction that started for the sha3(name). You are thinking about `sealedBid` which doesn't have any event..

HashRegistrarSimplified.sol
alexvandesande: What do you meant by indexed?

HashRegistrarSimplified.sol
Arachnid: Mostly just that it would exert gradually increasing pressure to do something. But then, a rising price of ether is likely to do that too. Perhaps we could even do away with the hard deadline altogether?

HashRegistrarSimplified.sol
Arachnid: Event parameters can have the keyword 'indexed', which means they get logged as separate properties in the log bloom filter. That means people can filter for just the events they care about - those targeting a specific name, or a specific owner.

HashRegistrarSimplified.sol
```diff
@@ -0,0 +1,340 @@
...
+
+import 'ENS.sol';
+
+
+contract Deed {
+    /* 
+    The Deed is a contract intended simply to hold ether
+    It can be controlled only by the registrar and can only send ether back to the owner.
+    */
+    Registrar public registrar;
```
Arachnid: Actually, this can be type address; it never calls any methods on the registrar.

HashRegistrarSimplified.sol
alexvandesande: I think the real important hard deadline is the one that makes all names impossible to release after 8 years. This would definitely reflect on the price of names..

We probably won't need those deadlines as we will update stuff long before them. But if we need them, we'll be glad we added. Sometimes a brick wall is necessary to win from inertia.

HashRegistrarSimplified.sol
alexvandesande: done

PR #10, opened 2016-09-14
Updates to hash registrar

alexvandesande:
* Refund bids if the auction is declared invalid
* Adds 'previousOwner' to deed
* Makes sure that bid fees burned are always related to value not deposit
* Invalidator now splits 50%
* Adds more tests


HashRegistrarSimplified.sol
```diff
@@ -23,10 +23,15 @@ import 'interface.sol';
  * @dev The deed can be controlled only by the registrar and can only send ether back to the owner.
  */
 contract Deed {
+    /* 
+    The Deed is a contract intended simply to hold ether
+    It can be controlled only by the registrar and can only send ether back to the owner.
```
Arachnid: This replicates the comments a few lines above.

PR #12, opened 2016-09-29
Add natspec comments

mardlin:


HashRegistrarSimplified.sol
```diff
@@ -12,17 +12,20 @@ and all ether still locked after 8 years will become unreachable.
...
 The plan is to test the basic features and then move to a new contract in at most
 2 years, when some sort of renewal mechanism will be enabled.
-
 */
 
+
 import 'ENS.sol';
 
+/*
+The Deed is a contract intended simply to hold ether
```
Arachnid: Nit: This should have *s on the left of each line - and should probably be a part of the following comment block.

HashRegistrarSimplified.sol
```diff
@@ -164,17 +201,23 @@ contract Registrar {
...
-    /*
-    ## Start Auction for available hash
+    /* 
+    ## Start Auction for an Available hash
 
     Anyone can start an auction by sending an array of hashes that they want to bid for. 
     Arrays are sent so that someone can open up an auction for X dummy hashes when they 
     are only really interested in bidding for one. This will increase the cost for an 
-    attacker from simply bidding on all new auctions blindly. Dummy auctions that are 
+    attacker to simply bid blindely on all new auctions. Dummy auctions that are
```
Arachnid: s/blindely/blindly/

HashRegistrarSimplified.sol
```diff
@@ -164,17 +201,23 @@ contract Registrar {
...
     Anyone can start an auction by sending an array of hashes that they want to bid for. 
     Arrays are sent so that someone can open up an auction for X dummy hashes when they 
     are only really interested in bidding for one. This will increase the cost for an 
-    attacker from simply bidding on all new auctions blindly. Dummy auctions that are 
+    attacker to simply bid blindely on all new auctions. Dummy auctions that are 
     open but not bid on are closed after a week. 
-    */    
+    */
+    
+    /**
```
Arachnid: Please make this part of the previous comment block.

HashRegistrarSimplified.sol
```diff
@@ -189,28 +232,44 @@ contract Registrar {
...
     contains information about the bid, including the bidded hash, the bid amount, and a random 
     salt. Bids are not tied to any one auction until they are revealed. The value of the bid 
     itself can be masqueraded by changing the required period or sending more than what you are 
     bidding for. This is followed by a 24h reveal period. Bids revealed after this period will 
     be burned and the ether unrecoverable. Since this is an auction, it is expected that most 
-    public hashes, like known domains and common dictionary words, will have multiple bidders pushing the price up. 
+    public hashes, like known domains and common dictionary words, will have multiple bidders 
+    pushing the price up. 
     */ 
+    /**
```
Arachnid: Please make this part of the previous comment block (eg, the previous comments can become the detailed description).

HashRegistrarSimplified.sol
```diff
@@ -231,6 +295,11 @@ contract Registrar {
         bid.setOwner(_owner);
         entry h = entries[_hash];
 
+        /* 
+        A penalty is applied for submitting pointless bids. This prevents bids which
+        are never intended to be revealed during the reveal period, which could otherwise
+        appear falsely as an expression of interest.
```
Arachnid: The main threat is that a participant could use the threat of revealing a bid that's higher than the second-highest bid to extort the winner into paying them.

HashRegistrarSimplified.sol
```diff
@@ -266,17 +336,28 @@ contract Registrar {
         }
     }
     
+    /**
+     * @dev Cancel a bid you've submitted
```
Arachnid: Anyone can call this; not just the person who submitted the bid.

HashRegistrarSimplified.sol
```diff
@@ -266,17 +336,28 @@ contract Registrar {
...
         bid.setOwner(msg.sender);
         bid.closeDeed(5);
         sealedBids[seal] = Deed(0);
         BidRevealed(seal, 0, 0, 5);
     }
-    
+
+    /**
+     * @dev Finalize an auction after the registration date has passed
+     * @param _hash
```
Arachnid: + the hash of the name the auction is for.

HashRegistrarSimplified.sol
```diff
@@ -324,6 +408,12 @@ contract Registrar {
     Names on the simplified registrar can't be six letters or less. We are purposefully
     handicapping its usefulness as a way to force it into being restructured in a few years
     */
+    /*
```
Arachnid: Please integrate this with the previous comments block.

Arachnid (commented):
Thanks for doing this! I left a few comments on the comments.

HashRegistrarSimplified.sol
mardlin: Like this? 
```/**
     * ## Blind auction for the desired hash
     * 
     * Bids are sent by sending a message to the main contract with a hash and an amount. The hash 
     * contains information about the bid, including the bidded hash, the bid amount, and a random 
     * salt. Bids are not tied to any one auction until they are revealed. The value of the bid 
     * itself can be masqueraded by changing the required period or sending more than what you are 
     * bidding for. This is followed by a 24h reveal period. Bids revealed after this period will 
     * be burned and the ether unrecoverable. Since this is an auction, it is expected that most 
     * public hashes, like known domains and common dictionary words, will have multiple bidders 
     * pushing the price up. 
     *  
     * @dev Submit a new sealed bid
     * @param sealedBid A sealedBid, created by the shaBid function
     */
```

HashRegistrarSimplified.sol
mardlin: ie. I reveal a bid, but another bid is waiting until the last second to reveal, I may be tempted to up my bid?

HashRegistrarSimplified.sol
mardlin: True. What is preventing someone from seeing another bidders sealed bid and cancelling it?

HashRegistrarSimplified.sol
Arachnid: Yes, although I'd put @dev at the top in place of ##.

HashRegistrarSimplified.sol
Arachnid: No; suppose I win the auction for 10 ether, and the next bid is 5 ether; I would expect to pay 5 ether. If there's no penalty for failing to reveal bids, Susan could place a bid every 1 ether increment, then after the auction come along and say "I will reveal this bid for 9 ether unless you pay me 1 ether not to".

HashRegistrarSimplified.sol
Arachnid: The bid can't be cancelled until 2 weeks after the auction ends.

HashRegistrarSimplified.sol
mardlin: Ahhhhhh, right.

mardlin:
Thanks for the feedback @Arachnid!

Arachnid:
Thanks!

jonesk7734 (approved):
Tested working

PR #14, opened 2016-10-14
Remove a @param comment that describes a non-existing parameter

pirapira:
`destroyDeed()` function takes no arguments but there was a @param comment describing an argument.  This was caught by the `develop` version of `solc`.


Arachnid:
Thanks!

PR #27, opened 2016-11-30
An LLL implementation of the public resolver

zigguratt:


PublicResolver.lll
```diff
@@ -0,0 +1,246 @@
...
+  ;; @notice Checks that the caller is the node owner.
+  ;; @param node Check the owner of this node.
+
+  (def 'only-owner (node)
+    (seq
+
+      ;; Call ens.owner(node).
+      (mstore (+ call-data 0x00) (shift-left get-node-owner))
+      (mstore (+ call-data 0x04) node)
+      (call (- (gas) 1000) 0 @@ens-registry call-data 36 return-data 32)
```
Arachnid: You should actually be able to drop the gas subtraction; after the most recent HF, call gas is automatically capped at 63/64 of caller's gas.

PublicResolver.lll
```diff
@@ -0,0 +1,246 @@
...
+
+      ;; If the call failed or the caller isn't the current node owner,
+      ;; throw an exception.
+      (when (|| (= @return-code false) (!= @return-data (caller)))
```
Arachnid: Isn't return code put on the stack?

PublicResolver.lll
```diff
@@ -0,0 +1,246 @@
...
+      ;; @notice Returns true if specified node has specified record type.
+      ;; @param node The ENS node to query.
+      ;; @param kind The record type name, as specified in EIP137.
+      ;; @return True if this resolver has a record of the provided type on the
+      ;;         provided node.
+
+      (def 'node (calldataload 0x04))
+      (def 'kind (calldataload 0x24))
+
+      (function has-record-type
```
Arachnid: Given we've deprecated this, I'd just drop it from the lll implementation, personally.

PublicResolver.lll
```diff
@@ -0,0 +1,246 @@
...
+  (def 'has-record-type    0x41b9dc2b) ; has(bytes32,bytes32)
+  (def 'supports-interface 0x01ffc9a7) ; supportsInterface(bytes4)
+  (def 'get-node-address   0x3b3b57de) ; addr(bytes32)
+  (def 'set-node-address   0xd5fa2b00) ; setAddr(bytes32,address)
+  (def 'get-content-hash   0xd8389dc5) ; hash(bytes32)
+  (def 'set-content-hash   0xd7fa1007) ; setHash(bytes32,bytes32)
+  (def 'get-node-owner     0x02571be3) ; owner(bytes32)
+
+  ;; Jumping here causes an EVM error.
+  (def 'invalid-location 0x02)  ; Jumping here causes an EVM panic.
```
zigguratt: `panic` is currently broken because of recent changes in the assembler to support first-class functions (iirc). Once that's fixed I'll switch this to use `panic`. In the meantime, `(jump 0x02)` does essentially the same thing.

PublicResolver.lll
zigguratt: Ah, good! I guess I can change that everywhere, then. Good to know. I'll update this contract first.

PublicResolver.lll
zigguratt: Consider it dropped.

PublicResolver.lll
zigguratt: I think you missed a [bug fix](https://github.com/ethereum/ens/pull/27/commits/598de2b735a421dc610196a2467bf37fcb7e8fbf) that was committed after the code you're looking at. In that code you'll see
```
(mstore return-code (call (- (gas) 1000) 0 @@ens-registry
```

PublicResolver.lll
```diff
@@ -0,0 +1,225 @@
...
+  ;; Constant definitions.
+
+  ;; Booleans.
+  (def 'true 1)
+  (def 'false 0)
+
+  ;; Memory layout.
+  (def 'scratch-one 0x00) ; Scratch space.
+  (def 'scratch-two 0x20) ; More scratch space.
+  (def 'return-code 0x40) ; Boolean returned to calling contract.
```
Arachnid: This seems to be an orphan now.

zigguratt: `scratch-two` didn't end up being used either. I'll remove both `return-code`and `scratch-two` and then renumber everything.

PR #35, opened 2017-01-06
Big rewrite of hash register to be explicitly state-machine based

Arachnid:


HashRegistrarSimplified.sol
```diff
@@ -118,19 +118,58 @@ contract Registrar {
...
         uint value;
         uint highestBid;
     }
 
+    // State transitions for names:
+    //   Open -> Auction (startAuction)
+    //   Auction -> Reveal
+    //   Reveal -> Owned
+    //   Reveal -> Open (if nobody bid)
+    //   Owned -> Forbidden (invalidateName)
```
alexvandesande: Shouldn't you be able to go from any state into forbidden?

HashRegistrarSimplified.sol
```diff
@@ -286,30 +318,23 @@ contract Registrar {
...
-            */
-            bid.closeDeed(10);
+        bid.setBalance(actualValue);
+
+        var auctionState = state(_hash);
+        if(auctionState == Mode.Owned) {
+            // Too late! Bidder loses their bid.
+            bid.closeDeed(0);
+            BidRevealed(_hash, _owner, actualValue, 1);
+        } else if(auctionState != Mode.Reveal) {
```
alexvandesande: Should you not be able to reveal it prematurely? You should not do it, but I'm not sure we should forbid it: 

The issues with revealing too soon is that you someone might overbid you, or worse, add a second-place bid in purpose just so they can make you spend unnecessary funds (or maybe blackmail you into it)

But the issue with revealing too late is that you lose 100% of the funds and you lose the opportunity to buy the name. This is a much worse scenario.

In the other hand, by not forbidding, it might make users get in the habit of premature revealing it on the app because they're impatient, which is worse.

HashRegistrarSimplified.sol
```diff
@@ -412,12 +434,11 @@ contract Registrar {
      * @param unhashedName An invalid name to search for in the registry.
      * 
      */
-    function invalidateName(string unhashedName) {
+    function invalidateName(string unhashedName) inState(sha3(unhashedName), Mode.Owned) {
```
alexvandesande: Again, if a name could be invalidated before auctions take place, then bids would be returned.

HashRegistrarSimplified.sol
```diff
@@ -286,30 +318,23 @@ contract Registrar {
...
-            /* 
-            A penalty (99%) is applied for submitting unrevealed bids, which could otherwise
-            be used as a threat of revealing a bid higher than the second-highest 
-            bid, to extort the winner into paying them.
-            */
-            bid.closeDeed(10);
+        bid.setBalance(actualValue);
+
+        var auctionState = state(_hash);
+        if(auctionState == Mode.Owned) {
```
alexvandesande: If we allow auctions to be invalidated at any point, then we could add a `if invalid` that would refund 100% of bids

HashRegistrarSimplified.sol
```diff
@@ -427,6 +448,7 @@ contract Registrar {
             h.deed.closeDeed(1000);
         }
         HashInvalidated(hash, unhashedName, h.value, now);
+        h.deed = Deed(0);
```
alexvandesande: Interesting solution!

test/simplehashregistrar_test.js
```diff
@@ -217,14 +218,10 @@ describe('SimpleHashRegistrar', function() {
 			function(done) {
 				async.each(bidData, function(bid, done) {
 					web3.eth.getBalance(bid.account, function(err, balance){
-						// Sleep sort is Best sort
-						// setTimeout(function() {
```
alexvandesande: ops totally forgot about this comment! :}

HashRegistrarSimplified.sol
Arachnid: Doing so complicates the bidding process a lot - so with this PR you can only forbid auctions that have finished. We could potentially add an Open -> Forbidden transition too.

HashRegistrarSimplified.sol
Arachnid: See above - with this PR, at least, you can't invalidate during an auction.

HashRegistrarSimplified.sol
Arachnid: This is another consequence of simplifying the code to be explicitly state-based; allowing reveal in multiple states would complicate the code. I don't think there's a compelling reason to complicate the code to allow early reveals.

PR #38, opened 2017-01-25
Creates an upgrade path

alexvandesande:
Creates an upgrade path from the old to a new registrar. 

Wasn’t able to test it yet

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -168,9 +168,9 @@ contract Registrar {
             if(entry.highestBid == 0 && now < registryCreated + upgradeDeadline) {
                 // If it seems open, check the previous registrar 
                 var (previousStatus, , previousRegistration, , ) = previousRegistrar.entries(_hash);
-                // Only valid one week after this registry was started
+                // Only valid one week after this registry was started (maybe replace with .eth ownership check?)
                 if (previousRegistration < registryCreated + auctionLength) {
-                    return previousStatus;
+                    return Mode(previousStatus);
                 } else {
                     return Mode.Open;
                 }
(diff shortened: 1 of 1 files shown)


HashRegistrarSimplified.sol
```diff
@@ -92,23 +92,49 @@ contract Deed {
...
+}
+
 /**
  * @title Registrar
  * @dev The registrar handles the auction process for each subnode of the node it owns.
  */
 contract Registrar {
     AbstractENS public ens;
     bytes32 public rootNode;
+    PreviousRegistrar public previousRegistrar;
```
Arachnid: Shouldn't this code be in a new contract (possibly one that extends this one)? It's not ideal to have a bunch of unused upgrade code in the original registrar that has nothing to upgrade from.

alexvandesande: Well I can put the full code of the ropsten registrar on a separate file, but this is supposed to be more general than just a ropstein 1.0 > ropstein 2.0. It could also be a migration path for the main net, where we could first publish a temporary code with a 4 week auction period and then after a month migrate to a 1 week auction, which is why I removed *initialAuctionPeriod*, and it could be the migration path to the permanent registrar and any other security upgrade necessary

Arachnid: I think we need to keep this contract as the one initially deployed anywhere the auction contract is used, and put a version that does upgrades in a separate contract - better yet, have it extend the base registrar with upgrade functionality.

I don't think we can do what you're suggesting with using upgrades to swap out the 4 week auction for a 1 week auction, because that would require everyone who registered on the initial contract to manually migrate their deposits over.

PR #39, opened 2017-02-20
Update tests due to solc 0.4.9 update

wulfraem:
As keys in result from `solc.compile(...)` now are prefixed with respective filenames, tests were failing on me.

For infos on update see:
solc-js 0.4.9 update: https://github.com/ethereum/solc-js/pull/82
Solidity 0.4.9 update: https://github.com/ethereum/solidity/releases/tag/v0.4.9


Arachnid:
Thanks!

PR #41, opened 2017-02-23
Adds a start date

alexvandesande:
If we want to be able to launch the ENS at a specific time in the future, it would be better to be able to deploy it and get everything set up before the auctions start. This gives an optional start parameter for the contract, to allow it only be activated on a future starting date


HashRegistrarSimplified.sol
```diff
@@ -175,10 +175,10 @@ contract Registrar {
      * @param _ens The address of the ENS
      * @param _rootNode The hash of the rootnode.
      */
-    function Registrar(address _ens, bytes32 _rootNode) {
+    function Registrar(address _ens, bytes32 _rootNode, uint _startDate) {
         ens = AbstractENS(_ens);
         rootNode = _rootNode;
-        registryCreated = now;
+        registryStarted = _startDate > 0 ? _startDate : now;
```
Arachnid: It probably makes more sense to just make this a mandatory parameter.

PR #43, opened 2017-03-07
Index sealed bids by sender, to prevent frontrunning

Arachnid:
By not keying sealed bids against the sender, we introduce a variety of frontrunning attacks, allowing attackers to invalidate others' bids. This patch fixes this by indexing each bid against the submitter, and making bids only revealable by the account that submitted them. Duplicate sealed bids are allowed as long as they come from different accounts.

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -101,7 +101,7 @@ contract Registrar {
     bytes32 public rootNode;
 
     mapping (bytes32 => entry) _entries;
-    mapping (bytes32 => Deed) public sealedBids;
+    mapping (address => mapping(bytes32 => Deed)) public sealedBids;
     
     enum Mode { Open, Auction, Owned, Forbidden, Reveal }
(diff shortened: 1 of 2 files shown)


alexvandesande:
I don't think that's a good solution. We want to have flexibility on the reveal, it's one of the most sensitive moments and it's when you are at the highest risk of losing funds, so I'd prefer if revealing can be done by anyone with the right information.

I prefer just using the value sent as part of the hash, so the worse you can do by front-running would be to pay for someone else's bid without any benefit to it.

Arachnid:
> I don't think that's a good solution. We want to have flexibility on the reveal, it's one of the most sensitive moments and it's when you are at the highest risk of losing funds, so I'd prefer if revealing can be done by anyone with the right information.

Doing so directly allows frontrunning on the reveal, since 'anyone with the right information' will immediately include anyone watching the transaction pool; they could then frontrun your reveal and take the name - and the deposit - for themselves.

PR #47, opened 2017-03-13
Remove unneeded check in the Registrar ctor

LefterisJP:
Removing a check for a `uint _startDate` being less than zero which will
always be false.

Other possible checks are if it's more than specific time (say 4 years)
in the past, but they are not necessarily needed.


LefterisJP:
Never mind. Misread the code.

Issue #49, opened 2017-03-14
HashRegistrarSimplified.sol: Would it be possible to win a bid without depositing sufficient ether?


swaldman:
I'm sure I'm just missing something dumb. So apologies in advance!

**TL; DR: Should the comparisons in [HashRegistrarSimplified.sol#L341](https://github.com/ethereum/ens/blob/28c66e5ff0b15f884c9318b33b9831b341ba84bd/HashRegistrarSimplified.sol#L341) and [HashRegistrarSimplified.sol#337](https://github.com/ethereum/ens/blob/28c66e5ff0b15f884c9318b33b9831b341ba84bd/HashRegistrarSimplified.sol#L337) and [HashRegistrarSimplified.sol#355](https://github.com/ethereum/ens/blob/28c66e5ff0b15f884c9318b33b9831b341ba84bd/HashRegistrarSimplified.sol#L337) be against `actualValue` rather than `_value`?**

Via [shaBid(bytes32 hash, address owner, uint value, bytes32 salt)](https://github.com/ethereum/ens/blob/28c66e5ff0b15f884c9318b33b9831b341ba84bd/HashRegistrarSimplified.sol#L284-L286), I can create a bid for associated with any value that I please. This value is part of the hash. I will always use this value to identify the bid. No ether is sent.

Via [newBid(bytes32 sealedBid)](https://github.com/ethereum/ens/blob/28c66e5ff0b15f884c9318b33b9831b341ba84bd/HashRegistrarSimplified.sol#L301-L310), I enter my bid into the auction. I may pay ether here, but there is no enforced relationship between the amount of ether I pay and the value I have declared when defining the bid. The value of my Deed will be defined by the amount I have paid, and the deed will own that amount.But this may not be the value I declared when defining the bid.

When I call `unsealBid`, I must identify the bid with its declared value `_value`, which may differ from the value of the `Deed` that represents it. The function will compute a quantity [actualValue](https://github.com/ethereum/ens/blob/28c66e5ff0b15f884c9318b33b9831b341ba84bd/HashRegistrarSimplified.sol#L326), which will be the minimum of the value I actually paid and the value declared. So, if I have underpaid, `actualValue` will be the value that I have actually paid.

[...]

PR #51, opened 2017-03-14
Underfunded bugfix

alexvandesande:
This fixes https://github.com/ethereum/ens/issues/49 and adds better tests to detect it


HashRegistrarSimplified.sol
```diff
@@ -334,11 +334,11 @@ contract Registrar {
         } else if(auctionState != Mode.Reveal) {
             // Invalid phase
             throw;
-        } else if (_value < minPrice) {
+        } else if (actualValue < minPrice) {
```
holiman: Actually... Why do we allow making a bid, which supposedly contains `_value` - the 'true' value, but then allow the actual bid value to be something else? It makes no sense (imho) to have such 'malleability'. I mean, there is a salt to provide entropy, so imho, the `_value` should be the sole source of **what was actually the bid value**. 

Therefore, I suggest that instead of 
```
uint actualValue = min(_value, bid.value())
```
We instead verify that `bid.value() >= _value`. And if not, consider the `bid` invalid. 

It's not a big issue, but I don't really like the ambiguity that is there right now.

Arachnid: We used to do that, but we didn't see any harm in this more lenient approach. The punishment for an invalid bid is losing it all, so we wanted to err on the side of not throwing away peoples' funds if there's no attack it enables.

PR #53, opened 2017-03-15
Avoid explicit type conversion.

chriseth:
Solidity allows some types in external functions than are not directly supported by the ABI. These include contract types, which are just interpreted as plain addreses as far as the ABI is concerned but proper type checking is done when functions are called via Solidity.


chriseth:
Found more occurences, will update.

chriseth:
Ok, updated.

Arachnid:
Thanks!

PR #57, opened 2017-03-22
Ignore calls to startAuction for auctions already running. Fixes #54

### LINKED ISSUE #54: Race condition in `startAuction`
`startAuction` and `startAuctions` throw if the auction was already in progress. This results in a race condition, where if two people open the same auction, one selected at random will get an error.

Instead, we should silently do nothing if the auction was already running, and only throw if the name is in a different state (owned, forbidden, reveal).

Arachnid:


HashRegistrarSimplified.sol
```diff
@@ -253,7 +253,11 @@ contract Registrar {
      *
      * @param _hash The hash to start an auction on
      */    
-    function startAuction(bytes32 _hash) inState(_hash, Mode.Open) registryOpen() {
+    function startAuction(bytes32 _hash) registryOpen() {
+        var mode = state(_hash);
+        if(mode == Mode.Auction) return;
+        if(mode != Mode.Open) throw;
```
holiman: Since this is not `payable`... why throw instead of `return`?

Arachnid: Because it's an error to try and open an auction that can't be opened (at least in my mind), and throwing reflects this. If this is being called by `startAuctions`, it also ensures that the call as a whole fails.

PR #58, opened 2017-03-24
Changes to registry interaction to improve upgradeability

Arachnid:
`finalizeAuction`, `releaseDeed`, and `invalidateName` all now work when the contract is the current registrar, by first checking this and skipping ENS changes if it's not. This means that these operations continue to function after the contract is replaced by a newer one, allowing in-progress auctions to complete normally.

This change means that the registrar can be replaced by a new one that need only care about transitioning completed auctions from the previous registrar. An upgrade process would go like this:
 1. Write a new registrar with an additional check in startAuction that prohibits starting any auction whose state in the old registrar is anything other than 'open'. Add a function to the new registrar that permits migrating a deed for a completed auction from the old registrar to the new one.
 2. Deploy the new registrar and set it as the active registrar in ENS.
 3. Anyone with an active auction on the old registrar at deployment time should complete the auction on the old registrar, then use the migration process to move the name to the new registrar in order to register it in ENS.
 4. New auctions should be started on the new registrar - the old one will not permit creation of any new auctions.

This means that the new registrar need only implement a restriction on starting auctions, and a migration process for completed auctions; it does not need to deal with auctions in progress at the time of deployment.

In addition, we allow users to release a name and recover their deposit as soon as the registrar is no longer current, rather than requiring them to wait a year if they don't want to upgrade.

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -391,8 +391,9 @@ contract Registrar {
 
         h.value =  max(h.value, minPrice);
 
-        // Assign the owner in ENS
-        ens.setSubnodeOwner(rootNode, _hash, h.deed.owner());
+        // Assign the owner in ENS, if we're still the registrar
+        if(ens.owner(rootNode) == address(this))
+            ens.setSubnodeOwner(rootNode, _hash, h.deed.owner());
 
         Deed deedContract = h.deed;
         deedContract.setBalance(h.value);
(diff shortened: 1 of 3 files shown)


HashRegistrarSimplified.sol
```diff
@@ -411,22 +412,24 @@ contract Registrar {
...
+     *      the name and get their ether back.
      * @param _hash The node to release
      */
     function releaseDeed(bytes32 _hash) onlyOwner(_hash) {
         entry h = _entries[_hash];
         Deed deedContract = h.deed;
-        if (now < h.registrationDate + 1 years 
-            || now > registryStarted + 8 years) throw;
+        if(now < h.registrationDate + 1 years && ens.owner(rootNode) == address(this)) throw;
+        if(now > registryStarted + 8 years) throw;
```
alexvandesande: We can remove this line

PR #59, opened 2017-03-24
Use Alpine instead of Ubuntu

nicksavers:
Saves about ~500MB storage and 182MB in network download if you don't have the `ubuntu:latest` image already. It costs you 5MB in network traffic if you already do.
https://www.brianchristner.io/docker-image-base-os-size-comparison/


Arachnid:
Thanks! I didn't worry overly since it's a oneoff task few people will run, but since you've already made the effort to improve it...

PR #65, opened 2017-03-25
Invalidating non-finalised auctions should only cost the bidder half their bid, not half their maximum

Arachnid:
Depends on #58.

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -455,7 +455,8 @@ contract Registrar {
         if(address(h.deed) != 0) {
             // Reward the discoverer with 50% of the deed
             // The previous owner gets 50%
-            h.deed.setBalance(h.deed.value()/2);
+            h.value = max(h.value, minPrice);
+            h.deed.setBalance(h.value/2);
             h.deed.setOwner(msg.sender);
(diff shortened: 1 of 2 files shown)


HashRegistrarSimplified.sol
```diff
@@ -451,7 +451,8 @@ contract Registrar {
         if(address(h.deed) != 0) {
             // Reward the discoverer with 50% of the deed
             // The previous owner gets 50%
-            h.deed.setBalance(h.deed.value()/2);
+            h.value = max(h.value, minPrice);
```
holiman: When would `minPrice` be greater than `h.value` ? 
And if so, can it be guaranteed that `h.deed.balance` is also larger than `h.value/2` ? 
Because if not, then `setBalance` could fail, and it wouldn't be possible to invalidate the name....

Arachnid: `minPrice` is greater than `h.value` if only one bid was received - eg, in the case of an uncontested auction. 

The deed's balance will always be greater than or equal to `h.value`, because when a winning bid is revealed, the `actualValue` is set to the minimum of the deed's balance and the declared bid. `h.value` is then set to `actualValue`.

PR #66, opened 2017-03-25
Allow unowned names to be zeroed out by anyone

Arachnid:
Also zeroes out resolver fields automatically when names are invalidated or released.

Depends on #65.

### DIFF
--- a/contracts/HashRegistrarSimplified.sol
+++ b/contracts/HashRegistrarSimplified.sol
@@ -440,8 +440,8 @@ contract Registrar {
         h.value = 0;
         h.highestBid = 0;
         h.deed = Deed(0);
-        
-        trySetSubnodeOwner(_hash, 0);
+
+        _tryEraseSingleNode(_hash);
         deedContract.closeDeed(1000);
         HashReleased(_hash, h.value);        
     }
@@ -460,7 +460,7 @@ contract Registrar {
 
         entry h = _entries[hash];
 
-        trySetSubnodeOwner(hash, 0);
+        _tryEraseSingleNode(hash);
 
         if(address(h.deed) != 0) {
             // Reward the discoverer with 50% of the deed
@@ -472,6 +472,45 @@ contract Registrar {
         }
         HashInvalidated(hash, unhashedName, h.value, h.registrationDate);
         h.deed = Deed(0);
+    }
+
+    /**
+     * @dev Allows anyone to delete the owner and resolver records for a (subdomain of) a
+     *      name that is not currently owned in the registrar. If passing, eg, 'foo.bar.eth',
+     *      the owner and resolver fields on 'foo.bar.eth' and 'bar.eth' will all be cleared.
+     * @param labels A series of label hashes identifying the name to zero out, rooted at the
+     *        registrar's root. Must contain at least one element. For instance, to zero 
+     *        'foo.bar.eth' on a registrar that owns '.eth', pass an array containing
+     *        [sha3('foo'), sha3('bar')].
+     */
+    function eraseNode(bytes32[] labels) {
+        if(labels.length == 0) throw;
+        if(state(labels[labels.length - 1]) == Mode.Owned) throw;
+
+        _eraseNodeHierarchy(labels.length - 1, labels, rootNode);
+    }
+
+    function _tryEraseSingleNode(bytes32 label) internal {
+        if(ens.owner(rootNode) == address(this)) {
+            ens.setSubnodeOwner(rootNode, label, address(this));
+            var node = sha3(rootNode, label);
+            ens.setResolver(node, 0);
+            ens.setOwner(node, 0);
+        }
+    }
+
+    function _eraseNodeHierarchy(uint idx, bytes32[] labels, bytes32 node) internal {
+        // Take ownership of the node
+        ens.setSubnodeOwner(node, labels[idx], address(this));
+        node = sha3(node, labels[idx]);
+        
+        // Recurse if there's more labels
+        if(idx > 0)
+            _eraseNodeHierarchy(idx - 1, labels, node);
+
+        // Erase the resolver and owner records
+        ens.setResolver(node, 0);
+        ens.setOwner(node, 0);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


HashRegistrarSimplified.sol
```diff
@@ -461,6 +468,36 @@ contract Registrar {
...
+    }
+
+    function _eraseNode(uint idx, bytes32[] labels, bytes32 node) internal {
+        // Take ownership of the node
+        ens.setSubnodeOwner(node, labels[idx], address(this));
+        node = sha3(node, labels[idx]);
+        
+        // Recurse if there's more labels
+        if(idx > 0)
+            _eraseNode(idx - 1, labels, node);
```
holiman: Hm. So what happens if I create (through multiple calls over a period of time)  a name like

`1.2.3.4.5.6.7.8........1024.foo.bar.eth`. Nobody can delete 1024 levels recursively. Is it possible to call `_eraseNode` on some lower level than the top level, to cut the recursion into several steps?

Arachnid: I did consider that; frankly, I'm happy for you to keep your ridiculously long domain, even if the 2LD is 'too short'.

There's no way I'm aware of to shortcut this without adding lots of complexity (such as storing intermediate values in contract storage).

holiman: Does not that mean I get to keep `bar.eth` also ?

Arachnid: No, `bar.eth` gets zeroed out when it gets invalidated, or when someone calls this method on `bar.eth`.

HashRegistrarSimplified.sol
```diff
@@ -460,6 +464,43 @@ contract Registrar {
...
+     *        [sha3('foo'), sha3('bar')].
+     */
+    function eraseNode(bytes32[] labels) {
+        if(labels.length == 0) throw;
+        if(state(labels[labels.length - 1]) == Mode.Owned) throw;
+
+        _eraseNode(labels.length - 1, labels, rootNode);
+    }
+
+    function _eraseNode(bytes32 label) {
```
pipermerriam: Is this missing an `internal` visibility specifier?

Arachnid: Yes; thank you.

HashRegistrarSimplified.sol
```diff
@@ -460,6 +464,43 @@ contract Registrar {
...
     /**
+     * @dev Allows anyone to delete the owner and resolver records for a (subdomain of) a
+     *      name that is not currently owned in the registrar. If passing, eg, 'foo.bar.eth',
+     *      the owner and resolver fields on 'foo.bar.eth' and 'bar.eth' will all be cleared.
+     * @param labels A series of label hashes identifying the name to zero out, rooted at the
+     *        registrar's root. Must contain at least one element. For instance, to zero 
+     *        'foo.bar.eth' on a registrar that owns '.eth', pass an array containing
+     *        [sha3('foo'), sha3('bar')].
+     */
+    function eraseNode(bytes32[] labels) {
```
Arachnid: This wouldn't work because `setOwner` would undo the work of `setSubnodeOwner`; the second time around the loop, your call to `setSubnodeOwner` would fail because you don't own the parent node. All the `setOwner` calls need to be executed after all the `setSubnodeOwner` calls.

Arachnid: Agreed on the internal functions, by the way - but I couldn't think of better names. Suggestions very welcome. :)

HashRegistrarSimplified.sol
```diff
@@ -460,6 +464,43 @@ contract Registrar {
...
+        if(labels.length == 0) throw;
+        if(state(labels[labels.length - 1]) == Mode.Owned) throw;
+
+        _eraseNode(labels.length - 1, labels, rootNode);
+    }
+
+    function _eraseNode(bytes32 label) internal {
```
pipermerriam: suggested rename `function _eraseSingleNode`

HashRegistrarSimplified.sol
```diff
@@ -460,6 +464,43 @@ contract Registrar {
...
+        ens.setSubnodeOwner(rootNode, label, address(this));
+        var node = sha3(rootNode, label);
+        ens.setResolver(node, 0);
+        ens.setOwner(node, 0);
+    }
+
+    function _eraseNode(uint idx, bytes32[] labels, bytes32 node) internal {
```
pipermerriam: suggested rename `function _recursiveEraseManyNodes`

Arachnid: Maybe just `eraseNodeHierarchy`? The fact it does it recursively isn't really important to the caller.

HashRegistrarSimplified.sol
```diff
@@ -460,6 +464,43 @@ contract Registrar {
...
+    }
+
+    function _eraseNodeHierarchy(uint idx, bytes32[] labels, bytes32 node) internal {
+        // Take ownership of the node
+        ens.setSubnodeOwner(node, labels[idx], address(this));
+        node = sha3(node, labels[idx]);
+        
+        // Recurse if there's more labels
+        if(idx > 0)
+            _eraseNodeHierarchy(idx - 1, labels, node);
```
holiman: I'd be interested to know, at 4.7M gas, how 'deep' can this function reach, at a maximum? I wonder if it's in the tens or the hundreds.. The SSTORE's must be quite costly, more so than the SHAs. My back-of-the-envelope calculation suggests ~100 levels.

Arachnid: Based on running a few tests with shorter series, yes, about 115 levels in the worst case (they're all already zero).

HashRegistrarSimplified.sol
```diff
@@ -460,6 +464,43 @@ contract Registrar {
...
+    function _eraseSingleNode(bytes32 label) internal {
+        ens.setSubnodeOwner(rootNode, label, address(this));
+        var node = sha3(rootNode, label);
+        ens.setResolver(node, 0);
+        ens.setOwner(node, 0);
+    }
+
+    function _eraseNodeHierarchy(uint idx, bytes32[] labels, bytes32 node) internal {
+        // Take ownership of the node
+        ens.setSubnodeOwner(node, labels[idx], address(this));
```
holiman: How come you're setting `this` as owner, instead of zeroing it ?

Arachnid: Because it needs to be the owner in order to change the resolver, or take ownership of the next sub-node.

Arachnid:
Ping @alexvandesande @holiman; do we want to include this?

alexvandesande:
The tests are failing with me on the first test on this branch. Have you updated it recently? Also, needs merging from master.

Arachnid:
@alexvandesande I've rebased this against the current master.

PR #70, opened 2017-03-26
Wrap in describe the allowed and forbidden tests on transferring ownership

xavierlepretre:
I was thinking that it would be neater to wrap the passing and failing tests in their own `describe`. By showing them together, and differing by a single account, it puts forward the difference that makes it pass or fail.
Is that the kind of change you welcome?

### DIFF
--- a/test/fifsregistrar_test.js
+++ b/test/fifsregistrar_test.js
@@ -18,33 +18,34 @@ describe('FIFSRegistrar', function() {
 	var registrar = null;
 	var ens = null;
 
-
 	before(function() {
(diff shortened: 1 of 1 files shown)


Arachnid:
Yes, this is a great idea, and should clean up the tests a lot. A couple of minor points:

 1. Please try and keep whitespace changes to their own PR; they make it a lot harder to review code if they're interspersed with significant changes.
 2. You might have noticed that newer tests use promises, because they're way more compact and easier to read. You don't _have_ to convert anything you're refactoring to promises, but it'd be nice...

PR #71, opened 2017-03-26
Added elements that are needed for my fresh install to run tests.

xavierlepretre:
For the record. YMMV.


Arachnid:
Thanks! I could have sworn I added the bluebird dep.

PR #72, opened 2017-03-26
Promisified ReverseRegistrar tests.

xavierlepretre:
Is this the kind of promisifying you are after?

### DIFF
--- a/test/reverseregistrar_test.js
+++ b/test/reverseregistrar_test.js
@@ -50,11 +50,13 @@ describe('ReverseRegistrar', function() {
                                 reject(err);
                             } else if (typeof contract.address !== "undefined") {
                                 resolve(contract);
+                            } else {
+                                // There is to hope that reject or resolve is called
                             }
                         });
                 });
             })
-            .then(function(contract) {
+            .then(contract => {
                 registrar = Promise.promisifyAll(contract);
                 return ens.setOwnerAsync(0, registrar.address, {from: accounts[0]});
             });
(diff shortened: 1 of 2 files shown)


Arachnid:
Largely, yes! See the promisified tests for the auction registrar for an instance of how it can be done more compactly with ES6 function literals with, IMO, little loss in readability.

Does Mocha play nicely if you return a promise, though?

xavierlepretre:
I prefer the arrow notation too, let me update.

Mocha plays nicely with promises. Beside, if you try to use both `done()` and return a promise chain, as in:

```javascript
    it('calculates node hashes correctly', function(done) {
        this.slow(150);
        return registrar.nodeAsync(accounts[0])
            .then(function(hash) {
                assert.equal(hash, node);
                done();
            });
    });
```
then it complains with:

```
  1) ReverseRegistrar calculates node hashes correctly:
     Error: Resolution method is overspecified. Specify a callback *or* return a Promise; not both.
```

Arachnid:
Right, I was using `.asCallback(done)` on the promise, but if Mocha plays nice, this is neater.

test/reverseregistrar_test.js
```diff
@@ -28,47 +31,47 @@ describe('ReverseRegistrar', function() {
...
-            }],
-            done
-        );
+    beforeEach(function() {
+        this.timeout(4000);
+        return utils.deployENSAsync(accounts[0])
+            .then(function(_ens) {
+                ens = _ens;
+                return new Promise(function(resolve, reject) {
+                    web3.eth.contract(JSON.parse(registrarCode.interface)).new(
```
Arachnid: Shouldn't it be possible to just promisify `new` instead?

test/reverseregistrar_test.js
```diff
@@ -28,47 +31,47 @@ describe('ReverseRegistrar', function() {
...
-    it('calculates node hashes correctly', function(done) {
-        registrar.node(accounts[0], function(err, hash) {
-            assert.equal(err, null, err);
-            assert.equal(hash, node);
-            done();
-        });
+    it('calculates node hashes correctly', function() {
+        this.slow(150);
+        return registrar.nodeAsync(accounts[0])
+            .then(hash =>assert.equal(hash, node));
```
Arachnid: Nit: missing space

test/reverseregistrar_test.js
xavierlepretre: The "right" value to resolve is the second call, because the first has no address yet. So the `promisifyAll` does not work.
On the other hand, we could make a generic promisifier that passes on the second call only.

test/reverseregistrar_test.js
xavierlepretre: 😱

test/reverseregistrar_test.js
Arachnid: I think this would be good in future, but let's cross that bridge when we come to it.

PR #73, opened 2017-03-27
Ignore specific error types in the tests

### LINKED ISSUE #69: Invalid opcode instead of invalid jump
I run `npm test` and the `assert` here https://github.com/ethereum/ens/blob/f24df55d4c7378e46be7b49085d6e9041be310f7/test/ens_test.js#L73 fails with, in short:

```javascript
AssertionError: Error: VM Exception while executing transaction: invalid opcode
    at Object.callback (/home/vagrant/DAPPS/ens/test/ens_test.js:73:11)
```

My versions are:

* OS: vagrant-ubuntu-trusty-64
* NodeJs: v6.10.0
* Mocha: 3.2.0
* project's `node_modules` testrpc: 2.2.4
* project's `node_modules` solc: 0.4.10
* project's `node_modules` web3: 0.16.0

The test at line 53 passes.

yuanfeiz:
Also add helper function for conversion from days to secs.

Fixed #69

### DIFF
--- a/test/simplehashregistrar_test.js
+++ b/test/simplehashregistrar_test.js
@@ -23,6 +23,11 @@ function advanceTime(delay, done) {
 }
 var advanceTimeAsync = Promise.promisify(advanceTime);
 
+function daysInSec(days) {
+	return days * 24 * 60 * 60;
+}
+
+
 describe('SimpleHashRegistrar', function() {
 	var registrarABI = null;
 	var registrarBytecode = null;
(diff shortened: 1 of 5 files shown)


test/simplehashregistrar_test.js
```diff
@@ -122,7 +127,7 @@ describe('SimpleHashRegistrar', function() {
 			// Check a duplicate bid would throw
 			function(done) {
 				registrar.newBid.call(bid, {from: accounts[0], value: 2e18}, function(err, result) {
-					assert.ok(err.toString().indexOf(utils.INVALID_JUMP) != -1, err);
+					assert.ok(err, err);
```
Arachnid: Have you tested this? I would expect `assert.ok(x)` to fail if x is undefined.

test/simplehashregistrar_test.js
```diff
@@ -23,6 +23,11 @@ function advanceTime(delay, done) {
 }
 var advanceTimeAsync = Promise.promisify(advanceTime);
 
+function daysInSec(days) {
```
Arachnid: Good idea. You could probably just call this 'days'.

test/simplehashregistrar_test.js
yuanfeiz: Yep, I think this snippet of the test is to ensure duplicate bid gets throw. So `err` is expected to be not `undefined` so `assert.ok(err)` would go through.

Arachnid: Oops, quite right.

PR #81, opened 2017-03-27
Don't throw if send fails when invalidating a name

### LINKED ISSUE #75: Don't allow the owner of a name to interfere with invalidation
...by throwing in their receive function

Arachnid:
Fixes #75.


alexvandesande:
LGTM

PR #82, opened 2017-03-27
Simplify deed, so it takes its balance in the constructor

### LINKED ISSUE #67: Deed creation is not atomic
It seems that Deed's value is always supposed to be less or equal to Deed's balance.  Though, Deed creation, as it is implemented in Registrar smart contract, is not atomic.  Deed is first created with some value and zero balance (https://github.com/ethereum/ens/blob/master/HashRegistrarSimplified.sol#L305) and then its balance is resupplied with exact same value (https://github.com/ethereum/ens/blob/master/HashRegistrarSimplified.sol#L309).

Deed creation could be made atomic via payable constructor:

```
function Deed() payable {
    ...
    value = msg.value;
}
```

That could be called like this:

```
Deed newBid = (new Deed).value(msg.value)();
```

Such refactoring will make default function in Deed contract probably not needed.

Arachnid:
Fixes #67.

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -43,11 +43,11 @@ contract Deed {
         _;
     }
 
-    function Deed(uint _value) {
+    function Deed() payable {
         registrar = msg.sender;
         creationDate = now;
         active = true;
-        value = _value;
+        value = msg.value;
     }
         
     function setOwner(address newOwner) onlyRegistrar {
@@ -88,9 +88,6 @@ contract Deed {
         if(owner.send(this.balance))
             selfdestruct(burn);
     }
-
-    // The default function just receives an amount
-    function () payable {}
 }
 
 /**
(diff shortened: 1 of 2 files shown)


HashRegistrarSimplified.sol
```diff
@@ -305,11 +302,9 @@ contract Registrar {
         if (address(sealedBids[msg.sender][sealedBid]) > 0 ) throw;
         if (msg.value < minPrice) throw;
         // creates a new hash contract with the owner
-        Deed newBid = new Deed(msg.value);
+        Deed newBid = (new Deed).value(msg.value)();
```
alexvandesande: Didn't know this could be done, care to explain it?

Anyway, LGTM

Arachnid: It lets you send value along with a creation call. Turns out it's been in since at least Solidity 0.4.0!

PR #83, opened 2017-03-27
Suggested promisifying of a contract factory.

xavierlepretre:
Notice how it resolves to an array with both returned contracts.
Only `contracts[1]` has the address, presumably.

### DIFF
--- a/test/utils.js
+++ b/test/utils.js
@@ -66,18 +66,14 @@ module.exports = {
 	promisifyContractFactory: function(contractFactory) {
 		contractFactory.newAsync = function() {
 			var args = arguments;
-			var contracts = [];
 			return new Promise(function(resolve, reject) {
 				args[args.length] = function(err, contract) {
 					if (err) {
 						reject(err);
+					} else if (typeof contract.address !== "undefined") {
+						resolve(contract);
 					} else {
-						contracts.push(contract);
-						if (contracts.length > 1) {
-							// Notice how it resolves to an array with both returned contract.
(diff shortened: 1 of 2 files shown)


Arachnid:
The constructor callback is called twice, once when the transaction is accepted into the node's transaction pool, and once when it's mined. For these purposes, it'd be sufficient to ignore the first callback and only fulfill the promise on the second.

It might be more convenient to just take the ABI, too, and do the `web3.eth.contract` call in the function.

xavierlepretre:
I can change to only return on the second time. 
But you are saying we should go with a synchronous `.new`? Wouldn't that be a problem with a future Web3, or a NodeJs that forbids synchronous calls?

Arachnid:
No, not at all - definitely use the asynchronous new. But there's no need to return an array in the promise - just return the value passed to the callback once `contract.address` is populated.

test/utils.js
```diff
@@ -62,5 +62,24 @@ module.exports = {
 		});
 	},
 	web3: web3,
-	TestRPC: TestRPC
+	TestRPC: TestRPC,
+	promisifyContractFactory: function(contractFactory) {
+		contractFactory.newAsync = function() {
+			var args = arguments;
```
Arachnid: I think this is left over from the earlier version? No need to store args now.

xavierlepretre: It is necessary, otherwise, inside `function(resolve, reject)`, `arguments are set to `{ '0': [Function resolve], '1': [Function reject] }`.

PR #85, opened 2017-03-27
forbid high value names

alexvandesande:
(In progress, haven't added tests yet: this is controversial and might not get support from the community)

This would prevent any names being registered for higher than 1000 ether on the permanent registrar. It means that if two bids higher than 1000 are placed on any name (by the same bidder or not) then the name cannot be registered by anyone. The purpose is to have a softer launch of ENS, and not allow high value names until the next update (which could even be a minor update that would remove only that particular feature)

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -388,14 +388,21 @@ contract Registrar {
         entry h = _entries[_hash];
 
         h.value =  max(h.value, minPrice);
+        h.deed.setBalance(h.value, true);
 
-        // Assign the owner in ENS, if we're still the registrar
-        if(ens.owner(rootNode) == address(this))
-            ens.setSubnodeOwner(rootNode, _hash, h.deed.owner());
+        if (h.value >= 1000 ether) {
+            // Temporary registrar will not accept high value names yet
+            h.deed.closeDeed(995);
+            h.deed = Deed(0);
+            HashInvalidated(hash, unhashedName, h.value, h.registrationDate);
 
-        Deed deedContract = h.deed;
-        deedContract.setBalance(h.value, true);
-        HashRegistered(_hash, deedContract.owner(), h.value, h.registrationDate);
+        } else {
+            // Assign the owner in ENS, if we're still the registrar
+            if(ens.owner(rootNode) == address(this))
+                ens.setSubnodeOwner(rootNode, _hash, h.deed.owner());
+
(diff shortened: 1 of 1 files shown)


HashRegistrarSimplified.sol
```diff
@@ -388,14 +388,21 @@ contract Registrar {
...
         h.value =  max(h.value, minPrice);
+        h.deed.setBalance(h.value, true);
 
-        // Assign the owner in ENS, if we're still the registrar
-        if(ens.owner(rootNode) == address(this))
-            ens.setSubnodeOwner(rootNode, _hash, h.deed.owner());
+        if (h.value >= 1000 ether) {
+            // Temporary registrar will not accept high value names yet
+            h.deed.closeDeed(995);
+            h.deed = Deed(0);
```
Arachnid: This invalidation happens during an auction, whereas previously you could only invalidate a name after an auction had completed. Do we know how this will behave WRT other bids on the auction?

alexvandesande: The invalidate is a different function because it requires further information. Since this rule only rely on information known at finalize time, I don't see why not do it at this step.

Arachnid: No, I was asking how this will affect existing behaviour; previously you could only invalidate after an auction, while this makes a name invalid _during_ an auction. It looks like this will result in revealBid rejecting any attempts to reveal any later bids, therefore locking up and eventually throwing away everyone else's ether.

alexvandesande: >  this makes a name invalid during an auction.

How so? It can only be called by owner, during state owned, which is by definition after the auction has ended.

Arachnid: I'm talking about _this_ - the code we're reviewing - not `invalidateName`.

Arachnid: Oops. Somehow I misplaced this code in my head and thought we were dealing with the reveal phase.

HashRegistrarSimplified.sol
```diff
@@ -388,14 +388,21 @@ contract Registrar {
         entry h = _entries[_hash];
 
         h.value =  max(h.value, minPrice);
+        h.deed.setBalance(h.value, true);
```
Arachnid: This should probably go in the 'else' clause - no point updating the balance twice if the if evaluates to true.

alexvandesande:
Closed in favor of https://github.com/ethereum/ens/pull/105

PR #86, opened 2017-03-28
Implement startAuctionsAndBid convenience method

### LINKED ISSUE #55: Add `startAuctionsAndBid` method
Since most addresses execute these two transactions one after the other, the loss in privacy is minimal, and this would allow UIs to be cleaner, since the user only has to approve one operation.

Arachnid:
Fixes #55.

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -305,7 +305,20 @@ contract Registrar {
         Deed newBid = (new Deed).value(msg.value)();
         sealedBids[msg.sender][sealedBid] = newBid;
         NewBid(sealedBid, msg.sender, msg.value);
-    } 
+    }
+
+    /**
+     * @dev Start a set of auctions and bid on one of them
+     * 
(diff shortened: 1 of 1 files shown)


HashRegistrarSimplified.sol
```diff
@@ -305,7 +305,20 @@ contract Registrar {
...
+     * @dev Start a set of auctions and bid on one of them
+     * 
+     * This method functions identically to calling `startAuctions` followed by `newBid`,
+     * but all in one transaction.
+     * @param hashes A list of hashes to start auctions on.
+     * @param sealedBid A sealed bid for one of the auctions.
+     */
+    function startAuctionsAndBid(bytes32[] hashes, bytes32 sealedBid) payable {
+        startAuctions(hashes);
+        newBid(sealedBid);
```
alexvandesande: just to confirm, the msg.value will be used in the newBid function, correct? We don't have to explicitly forward the value to it.

Arachnid: That's right.

PR #87, opened 2017-03-28
slowly lower the min length

alexvandesande:
Classify this as another attempt of trying to make the launch softer and slower without severely crippling the ENS functionality:

* Lower the initial auction period to 2 weeks
* Make it so that the initial name length starts at 12
* Every month, automatically reduce that limit one letter, arriving at the current state of 7 letters or more after 6 months

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -140,7 +140,7 @@ contract Registrar {
         } else {
             if(entry.highestBid == 0) {
                 return Mode.Open;
-            } else if(entry.deed == Deed(0)) {
+            } else if(minLength(now) == 6 && entry.deed == Deed(0)) {
                 return Mode.Forbidden;
             } else {
                 return Mode.Owned;
@@ -444,6 +444,15 @@ contract Registrar {
     }  
 
     /**
+     * @dev Minimum length starts at 12 weeks reducing 1 letter every 4 weeks until min of 6
+     * @param timestamp The time for which you want to know the date.
+     */
+
+    function minLength(uint timestamp) constant returns (uint minLength) {
+        return uint minLength = max(6, 12 - (timestamp - registryStarted)/(4 weeks));
+    }
+
+    /**
      * @dev Submit a name 6 characters long or less. If it has been registered, 
      * the submitter will earn 50% of the deed value. We are purposefully
      * handicapping the simplified registrar as a way to force it into being restructured
@@ -452,8 +461,7 @@ contract Registrar {
      * 
      */
     function invalidateName(string unhashedName) inState(sha3(unhashedName), Mode.Owned) {
-        uint minLength = max(6, 12 - (now - registryStarted)/(4 weeks));
-        if (strlen(unhashedName) > minLength ) throw;
+        if (strlen(unhashedName) > minLength(now) ) throw;
         bytes32 hash = sha3(unhashedName);
         
         entry h = _entries[hash];
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -104,7 +104,7 @@ contract Registrar {
     enum Mode { Open, Auction, Owned, Forbidden, Reveal }
     uint32 constant auctionLength = 5 days;
     uint32 constant revealPeriod = 48 hours;
-    uint32 constant initialAuctionPeriod = 4 weeks;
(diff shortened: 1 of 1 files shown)


pipermerriam:
I would suggest just starting it at whatever upper limit you think is appropriate to slow start the platform and expose a protected method that the multisig can call to reduce the limit (still with some lower bound).

That way you aren't locked into a specific schedule.  Seems bad to hard code those values since *things change*.

Arachnid:
To bring in a comment I made out of band: This is insufficient as it stands, since anyone can invalidate a name once, and it will remain invalidated indefinitely. Thus, I could invalidate 'ethereum.eth' on day 1, and it would be unregisterable even once the min length is short enough that it should be allowed.

pipermerriam:
Some thoughts this:

1. Remove the `Forbidden` state.  My current understanding suggests this would allow someone to continually re-register an invalid name.  Given that invalidating it again should be trivial and there is a financial incentive to do so, this seems like it could be an acceptable compromise.
2. With the minimum no longer being a constant, entries should contain a copy of what the minimum was at the time the auction was started. Invalidation should be done against this number rather than the current number.

alexvandesande:
Yes, with Nick's comment I think this approach is rather unviable as it would require lots of rewriting. If we have to refactor the forbidden state, add a new function etc, then it seems it would be easier simply to deploy a new contract and upgrade everything you want to lower the length

Arachnid:
I think Piper's suggestion of just removing the forbidden state is a reasonable solution. And while I agree that shortening it makes more sense as a manual operation than a timed one, doing that would require introducing new fields - to track who's authorised to make that call.

alexvandesande:
I changed the code so that minLength is a getter and that forbidden mode only starts applies after it has reached the target of 6.

What's your opinion? Would this add too much complexity and micromanagement for the contract? Is it worth for the purpose of making the launch a softer and slower event?

HashRegistrarSimplified.sol
```diff
@@ -444,6 +444,15 @@ contract Registrar {
     }  
 
     /**
+     * @dev Minimum length starts at 12 weeks reducing 1 letter every 4 weeks until min of 6
+     * @param timestamp The time for which you want to know the date.
+     */
+
+    function minLength(uint timestamp) constant returns (uint minLength) {
+        return uint minLength = max(6, 12 - (timestamp - registryStarted)/(4 weeks));
```
Arachnid: A year after deployment (52 weeks), `12 - (timestamp - registryStarted) / (4 weeks)` will evaluate to a negative number, which will underflow, and the min domain length will become too long for practical use.

HashRegistrarSimplified.sol
```diff
@@ -140,7 +140,7 @@ contract Registrar {
         } else {
             if(entry.highestBid == 0) {
                 return Mode.Open;
-            } else if(entry.deed == Deed(0)) {
+            } else if(minLength(now) == 6 && entry.deed == Deed(0)) {
```
Arachnid: This means that any names invalidated during the first 6 months will still show up as owned until the 6 month point, at which point they all become forbidden at once - it still won't allow anyone to reregister those names.

Arachnid:
To answer your other questions: I think this could be made workable if it just ditched the 'invalid' state entirely, per Piper's suggestion. I also think that if it were to be implemented, it would make sense as he suggests to make the method callable by the keyholders, so they can adjust the schedule of length decreases to suit conditions.

I'm of two minds about whether it's a good idea overall. On the one hand, reducing the length in stages would help spread out the initial 'land rush' and make it more manageable for people to keep up with it. On the other hand, it adds complexity that I'm not sure is justified.

pipermerriam:
The more I've thought about this the more I think that this should just be done by deploying a new registrar configured for a lower length.  Benefits I see in this approach.

* Reduced registrar complexity.
* Forces our hand to figure out a good deed migration pattern

I do think it's worth it to figure out a way to start with a high limit and lower it.  Slowing down the land grab will reduce a lot of risk while ENS get's battle tested.

Arachnid:
I like this suggestion - it means that the registrar doesn't have to give special permissions to anyone. However, it would require some way to upgrade the registrar that preserves running auctions - and I don't see any way to do that short of a more or less complete rewrite to use Piper's suggested database abstraction.

alexvandesande:
Closed in favor of https://github.com/ethereum/ens/pull/105

PR #104, opened 2017-03-31
Remove explicit owner field from bids

Arachnid:
Another strawman PR intended to provoke discussion; this one removes the owner field from bids, which somewhat simplifies the code, and addresses the issue Martin identified with accidentally setting it to 0.

### DIFF
--- a/HashRegistrarSimplified.sol
+++ b/HashRegistrarSimplified.sol
@@ -43,7 +43,8 @@ contract Deed {
         _;
     }
 
-    function Deed() payable {
+    function Deed(address _owner) payable {
+        owner = _owner;
         registrar = msg.sender;
         creationDate = now;
         active = true;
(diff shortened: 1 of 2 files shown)


HashRegistrarSimplified.sol
```diff
@@ -341,14 +339,14 @@ contract Registrar {
         if(auctionState == Mode.Owned) {
             // Too late! Bidder loses their bid. Get's 0.5% back.
             bid.closeDeed(5);
-            BidRevealed(_hash, _owner, actualValue, 1);
+            BidRevealed(_hash, msg.sender, actualValue, 1);
```
alexvandesande: I wonder if logging the bid.owner wouldn't be a more useful information

Arachnid: These should be the same thing.

PR #105, opened 2017-03-31
scheduled soft launch

alexvandesande:
This creates a softer launch by creating a random date for each hash, between launch and 6 months after launch, in which the given name can only be opened after the set date.

After thirteen weeks all names are available.


HashRegistrarSimplified.sol
```diff
@@ -104,7 +104,8 @@ contract Registrar {
     enum Mode { Open, Auction, Owned, Forbidden, Reveal }
     uint32 constant auctionLength = 5 days;
     uint32 constant revealPeriod = 48 hours;
-    uint32 constant initialAuctionPeriod = 4 weeks;
+    uint32 constant initialAuctionPeriod = 2 weeks;
```
Arachnid: Do we need a longer initial period under this proposal? Removing it might simplify the whole thing.

HashRegistrarSimplified.sol
```diff
@@ -104,7 +104,8 @@ contract Registrar {
...
+    uint32 constant lengthOfSoftLaunch = 26 weeks;
```
Arachnid: 6 months seems a bit excessive, TBH.

HashRegistrarSimplified.sol
```diff
@@ -237,6 +238,27 @@ contract Registrar {
...
+     * @dev Creates a scheduled release for each name
+     * 
+     * Each name will be assigned a random date in which its auction 
+     * can be started, from 0 to 6 months
+     * 
+     * @param _hash The hash to start an auction on
+     * @param _timestamp The timestamp to query about
+     */
+     
+    function isAllowed(bytes32 _hash, uint _timestamp) constant returns (bool allowed){
```
Arachnid: Could we write the inverse function, which returns the date at which an auction will be allowed, and then write `isAllowed` based on that?

alexvandesande: Yes. I was only thinking either it should be done on the client side, but I guess a reversal wouldnt hurt here.

HashRegistrarSimplified.sol
alexvandesande: What would be a good time in your opinion?

Arachnid: Somewhere around 2 or 3 months seems sensible to me, personally.

HashRegistrarSimplified.sol
alexvandesande: The problem doing so is that it should be a rather simple calculation on the client side: `registryStarted + launchLength * myHash/0xFFF`, but it runs on two issues doing on the blockchain using uints: you can't do (launchLength * myHash) because it might go over the overflow limit, and you can't invert it and do (0xFFF / myHash) because it returns a non-integer.

Arachnid: We can do (at a loss of precision we don't care about):

    var result = ((hash / 2**128) * launchLength) / 2**128;

HashRegistrarSimplified.sol
alexvandesande: reduced to 3 months

HashRegistrarSimplified.sol
alexvandesande: good idea. Added.

holiman:
Won't it be a bit of a PITA if you start the 4 auctions to hide your real one, and then one of the non-important ones keep 'blocking' the one that I really want ?

alexvandesande:
@holiman it shouldn't be hard to create fake auctions using this schema. Since availability is based on the hash size, you could simply generate a fake hash smaller than the limit. Also, the preimage file is ordered by hash, so you can also choose one by selecting among the N first hashes

test/simplehashregistrar_test.js
```diff
@@ -102,35 +105,191 @@ describe('SimpleHashRegistrar', function() {
...
+				});
+			},
+			function(done) { 
+				registrar.getAllowedTime(web3.sha3('ethereum'), function(err, result){ 
+					// console.log('When is "ethereum" allowed?\t', new Date(Number(result)*1000)); 
+					// 'Ethereum' should be available in 30 days
+					assert.equal(Math.round((Number(result)-registryStarted)/days(1)), 30);
+					done();
+				});
+			},
```
yuanfeiz: It might be better to extract the availability tests to a separate test case.

contracts/HashRegistrarSimplified.sol
```diff
@@ -238,7 +242,38 @@ contract Registrar {
...
 
+    /** 
+     * @dev Returns expected launch date for hash
+     * 
+     * Returns the minimum date for opening the auction
+     * 
+     * @param _hash The hash to start an auction on
+     */
+    function getAllowedTime(bytes32 _hash) constant returns (uint timestamp) {
+        return registryStarted + launchLength*(uint(_hash)/2**128)/2**128;
```
yuanfeiz: How about using the left shift operators? I did a simple benchmark, the shift version costs a little bit less gas than the division-power version at 296gas vs. 335gas. Not sure if it matters for the `constant` functions.

image

PS: any concern for `launchLength*(uint(_hash)/2**256)`

contracts/HashRegistrarSimplified.sol
```diff
@@ -11,7 +11,7 @@ names cannot be six letters or shorter, new auctions will stop after 4 years.
 
 The plan is to test the basic features and then move to a new contract in at most
 2 years, when some sort of renewal mechanism will be enabled.
-*/
```
Arachnid: Trailing whitespace.

contracts/HashRegistrarSimplified.sol
```diff
@@ -238,7 +242,38 @@ contract Registrar {
         }
         return len;
     }
+    
+    /** 
+     * @dev Creates a scheduled release for each name
```
Arachnid: Suggest "Determines if a name is available for registration yet" instead

contracts/HashRegistrarSimplified.sol
```diff
@@ -238,7 +242,38 @@ contract Registrar {
...
+     * 
+     * Each name will be assigned a random date in which its auction 
+     * can be started, from 0 to 6 months
```
Arachnid: 0 to 13 weeks.

contracts/HashRegistrarSimplified.sol
```diff
@@ -238,7 +242,38 @@ contract Registrar {
...
+     * 
+     * @param _hash The hash to start an auction on
+     * @param _timestamp The timestamp to query about
+     */
+     
+    function isAllowed(bytes32 _hash, uint _timestamp) constant returns (bool allowed){
```
Arachnid: I think it'd be a lot cleaner if this function just called `getAllowedTime` and compare it to the current time.

contracts/HashRegistrarSimplified.sol
```diff
@@ -238,7 +242,38 @@ contract Registrar {
...
+        uint timeSpan = _timestamp - registryStarted;
+        uint limit = timeSpan < launchLength ? 
+            timeSpan * (upperLimit / launchLength)
+            : upperLimit;
+        
+        return uint(_hash) < limit;
+    }
 
+    /** 
+     * @dev Returns expected launch date for hash
```
Arachnid: `s/expected//`

contracts/HashRegistrarSimplified.sol
```diff
@@ -238,7 +242,38 @@ contract Registrar {
...
+     * 
+     * Returns the minimum date for opening the auction
```
Arachnid: Isn't this saying the same thing as the line above?

contracts/HashRegistrarSimplified.sol
```diff
@@ -253,12 +288,12 @@ contract Registrar {
     function startAuction(bytes32 _hash) registryOpen() {
         var mode = state(_hash);
         if(mode == Mode.Auction) return;
-        if(mode != Mode.Open) throw;
-
+        if(mode != Mode.Open || !isAllowed(_hash, now)) throw;
```
Arachnid: Won't `mode == Mode.Unavailable` mean that the first part of this if statement fails?

contracts/HashRegistrarSimplified.sol
```diff
@@ -132,8 +133,11 @@ contract Registrar {
     //   Owned -> Open (releaseDeed)
     function state(bytes32 _hash) constant returns (Mode) {
         var entry = _entries[_hash];
-        if(now < entry.registrationDate) {
-            if(now < entry.registrationDate - revealPeriod) {
+        
+        if(!isAllowed(_hash, now)) {
+            return Mode.Unavailable;
```
Arachnid: Can we reuse `Mode.Forbidden` here? I don't think anything will treat the two differently.

contracts/HashRegistrarSimplified.sol
```diff
@@ -253,12 +288,12 @@ contract Registrar {
     function startAuction(bytes32 _hash) registryOpen() {
         var mode = state(_hash);
         if(mode == Mode.Auction) return;
-        if(mode != Mode.Open) throw;
-
+        if(mode != Mode.Open || !isAllowed(_hash, now)) throw;
+        
         entry newAuction = _entries[_hash];
 
         // for the first month of the registry, make longer auctions
```
Arachnid: This comment needs deleting.

contracts/HashRegistrarSimplified.sol
```diff
@@ -381,7 +416,7 @@ contract Registrar {
         Deed bid = sealedBids[bidder][seal];
         // If the bid hasn't been revealed after any possible auction date, then close it
         if (address(bid) == 0
-            || now < bid.creationDate() + initialAuctionPeriod) throw;
+            || now < bid.creationDate() + launchLength) throw;
```
Arachnid: I think this should be `auctionLength + revealPeriod`, instead.

test/simplehashregistrar_test.js
```diff
@@ -88,9 +88,12 @@ describe('SimpleHashRegistrar', function() {
...
 	it('starts auctions', function(done) {
 		async.series([
-			function(done) {
+			function(done) { 
+				advanceTime(days(100), done); 
+			},		
+			function(done) { 
 				registrar.startAuction(web3.sha3('name'), {from: accounts[0]}, done);
-			},
+			},
```
Arachnid: Trailing whitespace.

test/simplehashregistrar_test.js
```diff
@@ -88,9 +88,12 @@ describe('SimpleHashRegistrar', function() {
 
 	it('starts auctions', function(done) {
 		async.series([
-			function(done) {
+			function(done) { 
+				advanceTime(days(100), done);
```
Arachnid: Can you export the 'launchlength' constant and use it here and elsewhere instead of 100 days?

contracts/HashRegistrarSimplified.sol
yuanfeiz: That was a silly question for the "PS", I aha the constant zero part.

test/simplehashregistrar_test.js
```diff
@@ -102,35 +105,191 @@ describe('SimpleHashRegistrar', function() {
...
-			// Advance to the 24th day
-			function(done) { advanceTime(days(22), done); },
+			// Advance 30 days
+			function(done) { advanceTime(days(30), done); },
+			// Advancing days only have an effect after a transaction
 			function(done) {
 				registrar.startAuction(web3.sha3('anothername'), {from: accounts[0]}, done);
-			},
-			// Check later auctions end 7 days after they start
+			},
```
Arachnid: Trailing whitespace.

test/simplehashregistrar_test.js
```diff
@@ -102,35 +105,191 @@ describe('SimpleHashRegistrar', function() {
...
-			function(done) { advanceTime(days(22), done); },
+			// Advance 30 days
+			function(done) { advanceTime(days(30), done); },
+			// Advancing days only have an effect after a transaction
 			function(done) {
 				registrar.startAuction(web3.sha3('anothername'), {from: accounts[0]}, done);
-			},
-			// Check later auctions end 7 days after they start
+			},		
+			// Check later auctions end 5 days after they start
```
Arachnid: Is this relevant any longer?

test/simplehashregistrar_test.js
```diff
@@ -102,35 +105,191 @@ describe('SimpleHashRegistrar', function() {
...
+					var expectedEnd = new Date().getTime() / 1000 + days(137);
 					assert.ok(Math.abs(result[2].toNumber() - expectedEnd) < 5); // registrationDate
 					done();
 				});
-			}
+			},
+			function(done) {
+				registrar.entries(web3.sha3('name'), function(err, result) {
+					assert.equal(err, null, err);
+					// Asserts that auctions that go unbidded will become open again
```
Arachnid: Isn't this tested elsewhere?

test/simplehashregistrar_test.js
```diff
@@ -102,35 +105,191 @@ describe('SimpleHashRegistrar', function() {
...
+					assert.equal(Number(result[0]), 0); // status == Open
+					done();
+				});
+			}			
 		], done);
 	});
 
+	it('launch starts slow with scheduled availability', function(done) {
+		var launchLength = 7862400;
```
Arachnid: This would be clearer as `days(7 * 13)`

test/simplehashregistrar_test.js
```diff
@@ -102,35 +105,191 @@ describe('SimpleHashRegistrar', function() {
...
+		var registryStarted = 0;
+		async.series([
+			function(done) { 
+				registrar.registryStarted(function(err, result){
```
Arachnid: You should `assert.equal(err, null, err)` in these functions.

test/simplehashregistrar_test.js
```diff
@@ -102,35 +105,191 @@ describe('SimpleHashRegistrar', function() {
...
+		async.series([
+			function(done) { 
+				registrar.registryStarted(function(err, result){ 
+					registryStarted = Number(result); 
+					done();
+				});
+			},
+			function(done) { 
+				registrar.getAllowedTime('0x00', function(err, result){ 
+					// console.log('When is "0x00" allowed?\t', Number(result), Number(result)-registryStarted, (Number(result)-registryStarted)/launchLength);
```
Arachnid: Can you delete these commented lines?

test/simplehashregistrar_test.js
```diff
@@ -202,6 +365,15 @@ describe('SimpleHashRegistrar', function() {
 			// Start an auction for 'name'
 			function(done) {
 				registrar.startAuction(web3.sha3('name'), {from: accounts[0]}, done);
+			},	
+			function(done) {
+				registrar.entries(web3.sha3('name'), function(err, result) {
+					assert.equal(err, null, err);
+					// console.log('\n\n status', result);
```
Arachnid: More comments.

test/simplehashregistrar_test.js
```diff
@@ -385,24 +575,46 @@ describe('SimpleHashRegistrar', function() {
 			// Reveal the second bid
 			function(done) {
 				let bid = bidData[1];
-				registrar.unsealBid(web3.sha3('cancelname'), bid.value, bid.salt, {from: bid.account}, function(err, txid) {
-					assert.equal(err, null, err);
-					console.log('\t Bid #2 revealed');
-					done();
+
+				registrar.sealedBids(bid.account, bid.sealedBid, function(err, result) {
+					// console.log('\t Bid #2', err, result);
```
Arachnid: Comments.

test/simplehashregistrar_test.js
```diff
@@ -385,24 +575,46 @@ describe('SimpleHashRegistrar', function() {
...
+					// Checks the bid exists
+					assert.notEqual(result, '0x0000000000000000000000000000000000000000');
+					
+					registrar.unsealBid(web3.sha3('cancelname'), bid.value, bid.salt, {from: bid.account}, function(err, txid) {
+						assert.equal(err, null, err);
+						console.log('\t Bid #2 revealed');
+						
+						registrar.sealedBids(bid.account, bid.sealedBid, function(err, result) {
```
Arachnid: Why are these nested, instead of being more items in the `async.series`?

test/simplehashregistrar_test.js
```diff
@@ -529,18 +778,23 @@ describe('SimpleHashRegistrar', function() {
 
 			// Release the deed
 			function(done) {
-				registrar.releaseDeed(web3.sha3('releasename'), {from: bid.account}, function(err, txid) {
-					assert.equal(err, null, err);
-					console.log("\t Deed released");
-					web3.eth.getBalance(bid.account, function(err, balance){
-						assert.ok(balance.toFixed() > bid.startingBalance);
-						done();
+				// console.log('balance', web3.fromWei(balance.toFixed(), 'ether'));
```
Arachnid: More commented code.

test/simplehashregistrar_test.js
```diff
@@ -464,46 +700,59 @@ describe('SimpleHashRegistrar', function() {
 
 	it('releases deed after one year', function(done) {
 		this.timeout(5000);
-		var bid = {description: 'A regular bid', account: accounts[0], value: 1.1e18, deposit: 2.0e18, salt: 1, expectedFee: 0.005 };
+		var bidData = [{description: 'Winning bid', account: accounts[0], value: 10.0e18, deposit: 10.0e18, salt: 1, expectedFee: 0.005 },
+					{description: 'Second bid', account: accounts[1], value: 5.0e18, deposit: 10.0e18, salt: 1, expectedFee: 0.005 }];
```
Arachnid: Why add a second bid here?

test/simplehashregistrar_test.js
```diff
@@ -657,30 +912,33 @@ describe('SimpleHashRegistrar', function() {
...
 		var sealedBid = null;
-		registrar.shaBidAsync(web3.sha3('name'), 1e18, 1)
+
+		advanceTimeAsync(days(91))
+			.then((done) => registrar.shaBidAsync(web3.sha3('name'), 1e18, 1))
 			.then((result) => {
 				sealedBid = result;
 				return registrar.newBidAsync(result, {from: accounts[0], value: 1e18});
-			})
+			})
```
Arachnid: Trailing whitespace.

contracts/HashRegistrarSimplified.sol
alexvandesande: The app will certainly treat them differently, as unavailable has a date in which it becomes available. Of course, it can be client logic, not contract logic

contracts/HashRegistrarSimplified.sol
alexvandesande: lol 
True. I made this function first and then had a hard time reverting it into the getAllowedTime and didn't realize I didn't need the original one anymore.

contracts/HashRegistrarSimplified.sol
Arachnid: Yes, fair enough, let's leave it as-is then - although I'd prefer a clearer name than 'Unavailable' (not that one comes to mind).

contracts/HashRegistrarSimplified.sol
alexvandesande: There's an edge case here: if I bid on an auction and do not reveal, but neither does anyone else, you can still reopen the auction reveal again and not lose the deposit. It's unintentional and I imagine could be useful for people who are trying to buy obscure names.

Also, if for some reason you create a bid before opening an auction, it means that for a short period of time, your bid could be invalidated.

Finally, if you forgot to bid, revealing it after the launch will get your extra deposit back (and reveal some more data to the market) so it's always better for users to reveal invalid bids than cancelling them and making the cancel period short, it creates a race.

It seems to me increasing the time for cancelling a bid would be tolerant of those mistakes and still allow cleaning of transactions on the future.

test/simplehashregistrar_test.js
alexvandesande: Launch length is not a public function. Is there a good reason to make it so?

test/simplehashregistrar_test.js
alexvandesande: don't see why not test it at least once, if the test is there

contracts/HashRegistrarSimplified.sol
Arachnid: If we want to give people a grace period, then, we should explicitly set that grace period - there's no reason it needs to be identical to the soft launch period.

test/simplehashregistrar_test.js
Arachnid: No, I mean the 'launchLength' JS variable you define later in a test. If that's public, then you could use that everywhere you're currently using `advanceTime(days(90))` or `advanceTime(days(100))`, which would make your intent clearer.

test/simplehashregistrar_test.js
Arachnid: There's no longer any reason in the code that the auction duration would depend on when you start it, though - so it seems to me that the test just adds clutter.

test/simplehashregistrar_test.js
alexvandesande: To make sure it will work with values higher than the minimum value

contracts/HashRegistrarSimplified.sol
alexvandesande: Unreleased? NotYetAvailable? NotLaunched?

Arachnid:
Not yet available sounds good.

contracts/HashRegistrarSimplified.sol
alexvandesande: Making it auctionLength + 2 weeks. (auction length already includes the reveal length

contracts/HashRegistrarSimplified.sol
alexvandesande: According to remix bitshifting is actually more expensive: 
```
getAllowedTime(bytes32): 449
getAllowedTimeBitshift(bytes32): 527  
```

But I'll let it this way because I learned something new and it's more elegant that way

alexvandesande:
@uzyn it prevents every single auction being ended up happening on the same time. Not only it reduces the chances of large scale damaging on any bugs discovered, it also allows the markets to learn about how auctions work and allow apps to adapt to it with less pressure.

alexvandesande:
@Arachnid yes, I only saw the new pattern after writing some tests. I prefer it too.

test/simplehashregistrar_test.js
```diff
@@ -382,27 +563,49 @@ describe('SimpleHashRegistrar', function() {
...
 					bid.sealedBid = result;
-					registrar.cancelBid(bid.account, bid.sealedBid, {from: bid.account}, function(err, txid) {
-						assert.notEqual(err, null, err);
-						registrar.sealedBids(bid.account, bid.sealedBid, function(err, result) {
-							assert.equal(err, null, err);
-							assert.equal(result, 0);
-							console.log('\t Bid #2 not cancelled post-reveal. Sealedbid removed');
-							done();
+
+					registrar.sealedBids(bid.account, bid.sealedBid, function(err, result) {
```
Arachnid: These would be a lot easier to read if they were in series rather than nested.

test/simplehashregistrar_test.js
```diff
@@ -724,7 +977,7 @@ describe('SimpleHashRegistrar', function() {
 				});
 			},
 			// Advance 26 days to the reveal period
```
Arachnid: 3 days, not 26

test/simplehashregistrar_test.js
```diff
@@ -782,7 +1039,7 @@ describe('SimpleHashRegistrar', function() {
...
-			function(done) { advanceTime(days(26) + 1, done); },
+			function(done) { advanceTime(days(3) + 1, done); },
```
Arachnid: 3 days

test/simplehashregistrar_test.js
```diff
@@ -886,7 +1147,7 @@ describe('SimpleHashRegistrar', function() {
 				});
 			},
 			// Advance 26 days to the reveal period
```
Arachnid: 3 days

test/simplehashregistrar_test.js
```diff
@@ -712,157 +712,45 @@ describe('SimpleHashRegistrar', function() {
...
+			.then((result) => {
+				sealedBid = result;
+				return registrar.newBidAsync(result, {from: accounts[1], value: 2e18});
+			})
+			.then((done) => advanceTimeAsync(days(3) + 1))
+			.then((done) => registrar.unsealBidAsync(web3.sha3('name'), 1e18, 1, {from: accounts[0]}))
+			.then((done) => registrar.unsealBidAsync(web3.sha3('name'), 2e18, 2, {from: accounts[1]}))
+			.then((done) => advanceTimeAsync(days(2) + 1))
+			.then((done) => registrar.finalizeAuctionAsync(web3.sha3('name'), {from: accounts[1]}))
+			.then((done) => advanceTimeAsync(days(366) + 1))
```
Arachnid: Can you readd tests for releasing the deed early, and releasing the deed as the wrong user?

test/simplehashregistrar_test.js
```diff
@@ -1685,13 +1483,14 @@ describe('SimpleHashRegistrar', function() {
...
-		registrar.startAuctionAsync(web3.sha3('longname'), {from: accounts[0]})
+		advanceTimeAsync(launchLength)
+			.then((done) => registrar.startAuctionAsync(web3.sha3('longname'), {from: accounts[0]}))
 			.then((done) => registrar.shaBidAsync(web3.sha3('longname'), accounts[0], 1e18, 1))
 			.then((result) => {
 				sealedBid = result;
 				return registrar.newBidAsync(result, {from: accounts[0], value: 1e18});
 			})
-			.then((done) => advanceTimeAsync(26 * 24 * 60 * 60 + 1))
+			.then((done) => advanceTimeAsync(3 * 24 * 60 * 60 + 1))
```
Arachnid: This could use the new `days` function.

yuanfeiz:
Awesome! This is a huge progess!

Issue #99, opened 2017-03-30
Simplify the deployment process for private networks


Arachnid:
Ideally, via a 'deployment contract' that takes care of creating all the necessary contracts, then selfdestructs (or sits there and returns the relevant addresses for you)

yuanfeiz:
What do you think of using truffle? The footprint is as trivial as follow the naming conventions and have several configuration files.

The scripts to wire up ENS and FIFSRegistrar would look like this:
```
const ENS = artifacts.require('./ens/ENS.sol')
const FIFSRegistrar = artifacts.require('./ens/FIFSRegistrar.sol')
const PublicResolver = artifacts.require('./ens/PublicResolver.sol')
const Web3 = require('web3');
const web3 = new Web3();
const namehash = require('eth-ens-namehash');

const rootNode = namehash('eth');

module.exports = (deployer, network, accounts) => {
  deployer.deploy(ENS)
    .then(() => {
      return deployer.deploy(FIFSRegistrar, ENS.address, rootNode);
    })
    .then(function() {
      ENS.at(ENS.address).setSubnodeOwner('0x0', web3.sha3('eth'), FIFSRegistrar.address);
    });


  deployer.deploy(PublicResolver)
}
```

I can have a try on this and send a pull request.

Arachnid:
I'm amenable to the idea if it doesn't require rewriting the whole repository to be truffle-compatible. A pull request would be great.

PR #106, opened 2017-04-01
Adjust project directory to conform with truffle setup

yyyyyyuanfei:
#99


Arachnid:
Please let us know when this is ready to review.

yuanfeiz:
@Arachnid Alright, it's ready to be reviewed. Still plenty "todos" like to use truffle for compiling sol before the tests, I'll leave them for further improvements. 

One thing I'm not sure is that this PR introduces a new dependency to [eth-ens-namehash](https://www.npmjs.com/package/eth-ens-namehash), is there a better way to ensure the canonicity over this core algorithm while avoid implementing it in multiple places?

yuanfeiz:
And sorry I was out for three days, I'll be more responsive later.

migrations/2_deploy_contracts.js
```diff
@@ -0,0 +1,61 @@
...
+    });
+}
+
+/**
+ * Deploy the ENS and HashRegistrar(Simplified)
+ *
+ * @param {Object} deployer truffle deployer helper
+ * @param {string} rootNode hashed name using namehash algorithm
+ */
+function deployAuctionRegistrar(deployer, rootNode) {
```
Arachnid: This accepts an argument `rootNode`, but always attempts to register the registrar at 'eth'. It should either not take that argument (and just hardcode 'eth'), or take a label hash (eg, `sha3('eth')`) and compute the root node from that.

Arachnid:
Thank you - this looks like a big improvement, and will also make testing the DApp easier. I'm fine with the dependency on eth-ens-namehash.

Just one outstanding comment and then I'd be happy to merge it.

migrations/2_deploy_contracts.js
yuanfeiz: Good catch, fixed.

Arachnid (approved):

Arachnid:
Thanks!

PR #138, opened 2017-04-20
Remove forbidden state, and improve test handling of expected failures

Arachnid:


contracts/HashRegistrarSimplified.sol
```diff
@@ -130,8 +130,7 @@ contract Registrar {
     //   Auction -> Reveal
     //   Reveal -> Owned
     //   Reveal -> Open (if nobody bid)
-    //   Owned -> Forbidden (invalidateName)
-    //   Owned -> Open (releaseDeed)
+    //   Owned -> Open (releaseDeed or invalidateName)
```
alexvandesande: Does it make sense to also change line 106: `enum Mode { Open, Auction, Owned, Forbidden, Reveal, NotYetAvailable }`?

Pro: removes redundant code
Con: might break any code that expected Reveal and NotYetAvailable to be numbers 4 and 5

test/simplehashregistrar_test.js
```diff
@@ -28,6 +28,10 @@ function days(numberOfDays) {
 	return numberOfDays * 24 * 60 * 60;
 }
 
+function assertIsContractError(err) {
```
alexvandesande: this is great!

contracts/HashRegistrarSimplified.sol
Arachnid: I left it in because I was concerned about exactly that con. We could rename it to something like 'OBSOLETE_FORBIDDEN'.

alexvandesande: then just leave it as is. I bet we'll eventually use that space again in the future

test/simplehashregistrar_test.js
```diff
@@ -28,6 +28,10 @@ function days(numberOfDays) {
...
+	return assert.ok(err.toString().indexOf("invalid JUMP") != -1, err);
```
veox: This line, in conjunction with `ens/node_modules/solc/soljson.js`, seems to have caused all CI build done since to fail.

Arachnid: Yes, I've just fixed it.

PR #139, opened 2017-04-20
Make the launch length public

alexvandesande:
building the app, I miss having this information queriable as it varies according to testnet/mainet


alexvandesande:
If we are to do any other code change (removing invalid status or customizable name length for instance) we should definitely republish on the testnet. Certainly this by itself doesn't require a relaunch, but if we are doing one, it could go in.

PR #142, opened 2017-04-26
Set soft launch length to 8 weeks

Arachnid:


contracts/HashRegistrarSimplified.sol
```diff
@@ -130,8 +130,7 @@ contract Registrar {
     //   Auction -> Reveal
     //   Reveal -> Owned
     //   Reveal -> Open (if nobody bid)
-    //   Owned -> Forbidden (invalidateName)
-    //   Owned -> Open (releaseDeed)
+    //   Owned -> Open (releaseDeed or invalidateName)
```
alexvandesande: This is part of an unmarked PR. Will merge that first

Issue #124, opened 2017-04-07
destroyDeed clean up


ethers:
`selfdestruct(owner);` should be more reliable at cleaning up the deed

than

https://github.com/ethereum/ens/blob/5c2167b9fc6290c180e3989a30b4f9f4e3213d5a/contracts/HashRegistrarSimplified.sol#L89-L90
```
if(owner.send(this.balance))
            selfdestruct(burn);
```

Arachnid:
As I mentioned in the two other bugs you opened on this same issue, this is intentional: we want to give the owner a chance to recover their funds if they had a broken fallback function, rather than losing them forever.

ethers:
```
        if(owner.send(this.balance))
            selfdestruct(burn);
```
is an unusual and narrow pattern that a comment about explicitly calling the `owner` fallback function with only enough gas to log an event is desired.  Example:

```
        // Invoke owner fallback function to allow owner to log an event if desired.
        // Owner should also be aware that setBalance (and other cases) can call the 
        // fallback function and somewhat distinguishing between the cases may be desired
        // (but must be done within 2300 gas).
        if(owner.send(this.balance))
            selfdestruct(burn);
```

Arachnid:
Fair enough, and good suggestion for the comment (although I don't think we need to say anything about distinguishing the reasons funds are sent). Would you be open to sending a PR?

PR #148, opened 2017-04-29
Explain indirect selfdestruct

ethers:
Explanation of why the code is not simply `selfdestruct(owner)` https://github.com/ethereum/ens/issues/124


Arachnid:
Thanks!

PR #154, opened 2017-05-04
Remove no longer used Forbidden enum value

ethers:


Arachnid:
We deliberately decided not to do this, in order to avoid disrupting existing code that has expectations about the value of 'Reveal'.

PR #164, opened 2017-05-05
Use pygments-lexer-solidity package for Solidity syntax highlighting

veox:
### Issue

The [Implementers](http://docs.ens.domains/en/latest/implementers.html) page has Solidity code, but no syntax highlighting for that.

### Solution

Add the [pygments-lexer-solidity PyPI package](https://pypi.python.org/pypi/pygments-lexer-solidity) (and a `requirements.txt` at that), and use that.

### Requires changes to ReadTheDocs project!

As described in their [FAQ answer](https://docs.readthedocs.io/en/latest/faq.html#my-project-isn-t-building-with-autodoc).

On the [Advanced project settings](https://readthedocs.org/dashboard/ens/advanced/) page:

* tick "Install your project inside a virtualenv";
* set "requirements file" to `docs/requirements.txt`;
* (recommended) further down, tick "Give the virtual environment access to the global site-packages dir".

The latter will re-use system-wide Sphinx and theme, instead of installing them separately in the virtualenv, speeding up the build a bit.

**There will be 1 failed build on RTD:**

* If the PR is merged first, then RTD won't understand the `solidity` code block.
* If the RTD project changes are made first, then `doc/requirements.txt` won't be available.

### Notes

* `pygments-lexer-solidity` sources: https://gitlab.com/veox/pygments-lexer-solidity
* Comparison: [before](http://docs.ens.domains/en/latest/implementers.html#writing-a-registrar) and [after](https://veox-ens.readthedocs.io/en/use-pygments-solidity-lexer/implementers.html#writing-a-registrar).

### DIFF
--- a/docs/implementers.rst
+++ b/docs/implementers.rst
@@ -9,7 +9,7 @@ Writing a resolver
 
 Resolvers are specified in EIP137_. A resolver must implement the following method:
 
-::
+.. code-block:: solidity
 
     function supportsInterface(bytes4 interfaceID) constant returns (bool)
 
@@ -33,7 +33,7 @@ Additionally, the `content()` interface is currently used as a defacto standard 
 
 For example, a simple resolver that supports only the `addr` type might look something like this:
 
-::
+.. code-block:: solidity
 
     contract SimpleResolver {
         function supportsInterface(bytes4 interfaceID) constant returns (bool) {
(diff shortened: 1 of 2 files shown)


veox:
CI fails for same reason all build recently: [an "invalid EVM opcode"](https://travis-ci.org/ethereum/ens/jobs/229116547#L788), possibly due to improperly set environment variable.

PR #173, opened 2017-05-07
Format

### LINKED ISSUE #172: Minor inconsistencies
The code has minor inconsistencies. Examples below.

```
        if (address(sealedBids[msg.sender][sealedBid]) > 0 ) throw;
        if (msg.value < minPrice) throw;
```

=> Inconsistent spacing before closing `if` bracket

```
    bool active;


    modifier onlyRegistrar {
```

=> Three new lines in a row. Why?

```
            // Bid too low or too late, refund 99.5%
            bid.closeDeed(995);
            BidRevealed(_hash, msg.sender, value, 0);
        } else if (value > h.highestBid) {
            // new winner
```

=> Inconsistent comment capitalisation

```
    // Logged when the resolver for a node changes.
    event NewResolver(bytes32 indexed node, address resolver);

    // Logged when the TTL of a node changes
    event NewTTL(bytes32 indexed node, uint64 ttl);
```

=> Inconsistent comment termination (use of `.` vs no `.`)

```
     * @dev After some time, or if we're no longer the registrar, the owner can release
     *      the name and get their ether back.
```

vs 

```
     * @dev Submit a name 6 characters long or less. If it has been registered,
     * the submitter will earn 50% of the deed value. We are purposefully
```

=> Inconsistent multiline `@dev` formatting

```
// Recurse if there's more labels
```

=> Recurse if there are more labels

```
        if(idx > 0)
            _eraseNodeHierarchy(idx - 1, labels, node);
```

=> `if` statement without brackets on two lines. Why?

yuanfeiz:
Resolved #172

### DIFF
--- a/contracts/HashRegistrarSimplified.sol
+++ b/contracts/HashRegistrarSimplified.sol
@@ -268,12 +268,12 @@ contract Registrar {
      * @dev Returns available date for hash
      * 
      * The available time from the `registryStarted` for a hash is proportional
-     * to its numberic value.
+     * to its numeric value.
      * 
      * @param _hash The hash to start an auction on
      */
     function getAllowedTime(bytes32 _hash) constant returns (uint timestamp) {
-        return registryStarted + ((launchLength * (uint(_hash) >> 128)) >>128);
+        return registryStarted + ((launchLength * (uint(_hash) >> 128)) >> 128);
         // Right shift operator: a >> b == a / 2**b
     }
 
--- a/contracts/HashRegistrarSimplified.sol
+++ b/contracts/HashRegistrarSimplified.sol
@@ -32,7 +32,6 @@ contract Deed {
     event DeedClosed();
     bool active;
 
-
     modifier onlyRegistrar {
         if (msg.sender != registrar) throw;
         _;
@@ -72,6 +71,7 @@ contract Deed {
 
     /**
      * @dev Close a deed and refund a specified fraction of the bid value
+     *
      * @param refundRatio The amount*1/1000 to refund
      */
     function closeDeed(uint refundRatio) onlyRegistrar onlyActive {
@@ -90,7 +90,7 @@ contract Deed {
         // Instead of selfdestruct(owner), invoke owner fallback function to allow
         // owner to log an event if desired; but owner should also be aware that
         // its fallback function can also be invoked by setBalance
-        if(owner.send(this.balance)) {
+        if (owner.send(this.balance)) {
             selfdestruct(burn);
         }
     }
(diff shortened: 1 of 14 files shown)


contracts/HashRegistrarSimplified.sol
```diff
@@ -251,50 +255,48 @@ contract Registrar {
...
-     
     function isAllowed(bytes32 _hash, uint _timestamp) constant returns (bool allowed){
         return _timestamp > getAllowedTime(_hash);
     }
 
     /** 
      * @dev Returns available date for hash
      * 
+     * The available time from the `registryStarted` for a hash is proportional
+     * to its numberic value.
```
yuanfeiz: Not sure if this explains the algorithm clearly?

contracts/HashRegistrarSimplified.sol
```diff
@@ -251,50 +255,48 @@ contract Registrar {
...
+     * 
      * @param _hash The hash to start an auction on
      */
     function getAllowedTime(bytes32 _hash) constant returns (uint timestamp) {
-        return registryStarted + (launchLength*(uint(_hash)>>128)>>128);
-        // right shift operator: a >> b == a / 2**b
+        return registryStarted + ((launchLength * (uint(_hash) >> 128)) >>128);
```
yuanfeiz: Add one more bracket to emphasis the priority.

contracts/HashRegistrarSimplified.sol
```diff
@@ -305,16 +307,24 @@ contract Registrar {
 
     /**
      * @dev Start multiple auctions for better anonymity
+     *
+     * Anyone can start an auction by sending an array of hashes that they want to bid for.
+     * Arrays are sent so that someone can open up an auction for X dummy hashes when they
+     * are only really interested in bidding for one. This will increase the cost for an
+     * attacker to simply bid blindly on all new auctions. Dummy auctions that are
+     * open but not bid on are closed after a week.
+     *
```
yuanfeiz: This information should be attached with this bulk version of `startAuctions` rather than the single version above.

contracts/HashRegistrarSimplified.sol
Arachnid: I think it's a reasonble description, but it's 'numeric', not 'numberic'.

contracts/HashRegistrarSimplified.sol
Arachnid: Should probably put spaces around the second right shift too.

migrations/2_deploy_contracts.js
```diff
@@ -7,7 +7,7 @@ const FIFSRegistrar = artifacts.require('./FIFSRegistrar.sol');
 // TODO: align the contract name with the source code file name.
 const Registrar = artifacts.require('./Registrar.sol');
 const web3 = new (require('web3'))();
-const namehash = require('eth-ens-namehash');
+const namehash = require('../node_modules/eth-ens-namehash');
```
Arachnid: I'd prefer we just added this as an explicit dependency.

yuanfeiz: It's currently an known issue for truffle having trouble to do it the regular way. But would be resolved in the next release.

PR #176, opened 2017-05-09
[Document]Fix highest bid value access index

yuanfeiz:
It has been changed to 4

```
function entries(bytes32 _hash) constant returns (Mode, address, uint, uint, uint) {
        entry h = _entries[_hash];
        return (state(_hash), h.deed, h.registrationDate, h.value, h.highestBid);
    }
```


Arachnid:
Thanks!

PR #192, opened 2017-05-18
Added text to PublicResolver.sol and test cases.

ricmoo:
I have only tested against `--network test`.

Please let me know if there is any issues or style you would like changed.

### DIFF
--- a/contracts/PublicResolver.sol
+++ b/contracts/PublicResolver.sol
@@ -20,7 +20,7 @@ contract PublicResolver {
     event NameChanged(bytes32 indexed node, string name);
     event ABIChanged(bytes32 indexed node, uint256 indexed contentType);
     event PubkeyChanged(bytes32 indexed node, bytes32 x, bytes32 y);
-    event TextChanged(bytes32 indexed node, string key, string value);
+    event TextChanged(bytes32 indexed node, string indexed indexedKey, string key);
 
     struct PublicKey {
         bytes32 x;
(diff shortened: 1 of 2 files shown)


contracts/PublicResolver.sol
```diff
@@ -13,12 +13,14 @@ contract PublicResolver {
...
     bytes4 constant ABI_INTERFACE_ID = 0x2203ab56;
     bytes4 constant PUBKEY_INTERFACE_ID = 0xc8690233;
+    bytes4 constant TEXT_INTERFACE_ID = 0x59d1d43c;
 
     event AddrChanged(bytes32 indexed node, address a);
     event ContentChanged(bytes32 indexed node, bytes32 hash);
     event NameChanged(bytes32 indexed node, string name);
     event ABIChanged(bytes32 indexed node, uint256 indexed contentType);
     event PubkeyChanged(bytes32 indexed node, bytes32 x, bytes32 y);
+    event TextChanged(bytes32 indexed node, string key, string value);
```
Arachnid: I don't think there's much point in logging the value here; strings are only logged hashed. Also, `key` should be indexed.

contracts/PublicResolver.sol
```diff
@@ -30,6 +32,7 @@ contract PublicResolver {
         bytes32 content;
         string name;
         PublicKey pubkey;
+        mapping (string => string) text;
```
Arachnid: Can you update this to be consistent with the spacing of the rest of the entries?

ricmoo:
I have made the recommended changes to style and the event.

contracts/PublicResolver.sol
ricmoo: As discussed on gitter, I've updated this to include the key indexed and non-indexed and removed value.

contracts/PublicResolver.sol
ricmoo: Done.

PR #193, opened 2017-05-19
Update registrar address?

swaldman:
I may (always) be missing something, but is the address for the ENS Registrar in the docs stale?


yuanfeiz:
I think the document is correct, the address is for the ENS not the registrar.

Arachnid:
@yuanfeiz is correct; the address is for ENS, not for the registrar. You can look up the registrar with `ens.owner(namehash('eth'))`.

swaldman:
ah, ok. thanks.

PR #196, opened 2017-05-24
Rewrite reverse registrar for better usability

Arachnid:
This PR is a rewrite of the reverse registrar.

Previously, the registrar would allow you to 'claim' a reverse record, assigning ownership to you. The usual flow for a user was this:

 1. Call `ReverseRegistrar.claim()`
 2. Call `ENS.setResolver()` to set a resolver
 3. Call `Resolver.setName()` to set the reverse name

This requires 3 transactions in the common case.

This PR adds new functionality to the reverse resolver, defining a simple default reverse registrar, and adding two new functions:

 - `claimWithResolver`, which assigns ownership of the reverse node after first setting a resolver record.
 - `setName`, which sets the resolver of the reverse node to the default reverse resolver, then updates the name record on it.

Thus, users wanting to set a custom resolver now only need two transactions instead of 3, and users just wanting to set a reverse name now need only a single transaction.

### DIFF
--- a/contracts/ReverseRegistrar.sol
+++ b/contracts/ReverseRegistrar.sol
@@ -1,21 +1,78 @@
-pragma solidity ^0.4.0;
+pragma solidity ^0.4.10;
 
-import './AbstractENS.sol';
+import "./AbstractENS.sol";
+
+contract Resolver {
+    function setName(bytes32 node, string name) public;
+}
+
+/**
+ * @dev Provides a default implementation of a resolver for reverse records,
+ * which permits only the owner to update it.
+ */
+contract DefaultReverseResolver is Resolver {
+    // namehash('addr.reverse')
+    bytes32 constant ADDR_REVERSE_NODE = 0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2;
+
+    AbstractENS public ens;
+    mapping(bytes32=>string) public name;
+    
+    /**
+     * @dev Constructor
+     * @param ensAddr The address of the ENS registry.
+     */
+    function DefaultReverseResolver(AbstractENS ensAddr) {
+        ens = ensAddr;
+
+        // Assign ownership of the reverse record to our deployer
+        var registrar = ReverseRegistrar(ens.owner(ADDR_REVERSE_NODE));
+        if(address(registrar) != 0) {
+            registrar.claim(msg.sender);
+        }
+    }
+
+    /**
+     * @dev Only permits calls by the reverse registrar.
+     * @param node The node permission is required for.
+     */
+    modifier owner_only(bytes32 node) {
+        require(msg.sender == ens.owner(node));
+        _;
+    }
+
+    /**
+     * @dev Sets the name for a node.
+     * @param node The node to update.
+     * @param _name The name to set.
+     */
+    function setName(bytes32 node, string _name) public owner_only(node) {
+        name[node] = _name;
+    }
+}
 
 contract ReverseRegistrar {
+    // namehash('addr.reverse')
+    bytes32 constant ADDR_REVERSE_NODE = 0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2;
+
     AbstractENS public ens;
-    bytes32 public rootNode;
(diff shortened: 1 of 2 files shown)


test/reverseregistrar_test.js
```diff
@@ -74,4 +97,48 @@ describe('ReverseRegistrar', function() {
...
+            .then(txHash => registrar.claimAsync(accounts[3], {from: accounts[0]}))
+            .then(txHash => ens.ownerAsync(node))
+            .then(result => assert.equal(result, accounts[3]))
+            .then(result => ens.resolverAsync(node))
+            .then(result => assert.equal(result, accounts[2]));
+    })
+
+    it('sets name records', function() {
+        this.slow(300);
+        return registrar.setNameAsync('testname', {from: accounts[0]})
```
veox: Could it be that CI is failing due to insufficient gas?..

contracts/ReverseRegistrar.sol
```diff
@@ -1,31 +1,117 @@
-pragma solidity ^0.4.0;
+pragma solidity ^0.4.10;
 
-import './AbstractENS.sol';
+import "AbstractENS.sol";
```
chriseth: Please try to use relative paths, they are much more portable.

contracts/ReverseRegistrar.sol
```diff
@@ -1,31 +1,117 @@
...
+    bytes32 constant ADDR_REVERSE_NODE = 0x91d1777781884d03a6757a803996e38de2a42967fb37eeaca72729271025a9e2;
+
+    AbstractENS ens;
+    mapping(bytes32=>string) public name;
+    
+    /**
+     * @dev Constructor
+     * @param ensAddr The address of the ENS registry.
+     */
+    function DefaultReverseResolver(address ensAddr) {
```
chriseth: You can already use `AbstractENS` as the type here instead of address, then you do not need an explicit conversion.

contracts/ReverseRegistrar.sol
```diff
@@ -1,31 +1,117 @@
...
+        ens = AbstractENS(ensAddr);
+    }
+
+    /**
+     * @dev Only permits calls by the reverse registrar.
+     * @param node The node permission is required for.
+     */
+    modifier owner_only(bytes32 node) {
+        if(msg.sender != ens.owner(node)) throw;
```
chriseth: `require(...)`?

test/reverseregistrar_test.js
veox: Could be the case - see [successful Travis build from my repo](https://travis-ci.org/veox/ens/builds/235608950), with an [arbitrarily high gas specified](https://github.com/veox/ens/commit/8e56257eb309ca84da40181081f055b5e487bf24).

contracts/ReverseRegistrar.sol
```diff
@@ -1,31 +1,117 @@
...
+    /**
+     * @dev Sets the `name()` record for the reverse ENS record associated with
+     * the calling account. First updates the resolver to the default reverse
+     * resolver if necessary.
+     * @param name The name to set for this address.
+     * @return The ENS node hash of the reverse record.
+     */
+    function setName(string name) returns (bytes32 node) {
+        node = claimWithResolver(this, defaultResolver);
+        defaultResolver.setName(node, name);
```
alexvandesande: where does it check that the sha of the name actually resolves to this node?

test/reverseregistrar_test.js
Arachnid: I tried everything but that, because it was saying "invalid opcode" not "insufficient gas". Thanks.

contracts/ReverseRegistrar.sol
Arachnid: It doesn't - you can set your reverse record to any name; callers have to check that it's consistent. Forcing this here wouldn't help, because someone could set the name to match, then update it later - or just set their own resolver that behaves however they want.

Arachnid:
PTAL.

contracts/ReverseRegistrar.sol
alexvandesande: I see, it makes sense.

PR #213, opened 2017-06-11
Expanded explanation returned value from .entries

jefflau:


docs/userguide.rst
```diff
@@ -57,13 +57,21 @@ Before placing a bid, you need to check if the name is available. Run this code
 
     ethRegistrar.entries(web3.sha3('name'))[0];
 
-If the returned value is `0`, the name is available, and not currently up for auction. If the returned value is `1`, the name is currently up for auction. If the returned value is `5`, that means that the 'soft launch' is in effect, and your name isn't yet available; you can check when it will be available for auction with:
+This will return a value with a total of 6 numbers between 0 and 5. The full solidity data structure for this can be viewed `here <https://github.com/ethereum/ens/blob/master/contracts/HashRegistrarSimplified.sol#L110>`_ in the Registrar contract. The numbers represent different 'states' a name is currently in.
```
Arachnid: This implies that the return is a 6-tuple, rather than a single number between 0-5. Can you rephrase?

PR #216, opened 2017-06-18
Correct all calls to eth.contract

foogunlana:
I noticed eth.contract breaks unless ```var eth = web3.eth``` earlier.
So I changed the eth.contract calls to web3.eth.contract calls so the code can be pasted directly in the console


foogunlana:
No problemo!

PR #228, opened 2017-07-23
docs: ens.setOwner() clarification + minor Sphinx warning clean-up

veox:
This PR aims to clarify that `ens.setOwner()` is not meant for full ownership transfer of names acquired through the `.eth` Registrar (as highligted in a recent gitter discussion).

The last commit 1e6e8f2 is not related, but got tacked on to quell Sphinx's warnings, as it's otherwise difficult to check changes locally.

See [doc example](https://veox-ens.readthedocs.io/en/doc-ens-setowner-note/userguide.html#transferring-a-name) in my RTD.

-----

Note: CI seems to fail in my repo. :/

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -179,7 +179,7 @@ Once called, the winning bidder will be refunded the difference between their bi
 If you are the winning bidder, congratulations!
 
 Managing Ownership
-----------------------
+------------------
 
 After finalizing, your account now owns both the name in ENS and the deed in the Auction Registrar.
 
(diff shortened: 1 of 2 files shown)


docs/userguide.rst
```diff
@@ -251,13 +253,19 @@ The above example configures 'somename.eth' to resolve to the address of your pr
 Transferring a name
 -------------------
 
-You can transfer ownership of a name you own in the ENS registry to someone else using `setOwner`:
+You can transfer ownership of a name you own in the ENS registry to another trusted account using `setOwner`:
```
Arachnid: This change seems superfluous. You can transfer ownership to any account, whether you trust them or not.

docs/userguide.rst
```diff
@@ -251,13 +253,19 @@ The above example configures 'somename.eth' to resolve to the address of your pr
...
 
 ::
 
     > ens.setOwner(namehash('somename.eth'), newOwner, {from: eth.accounts[0]});
 
-Note, however, that if the name was acquired through a registrar, such as through an auction described above, this will not transfer ownership of the locked bid! It will also not perform any administrative tasks that a registrar might want to do.
+This way, the bidding/renewal account ``eth.accounts[0]`` can be kept separate from the day-to-day control account ``newOwner``.
```
Arachnid: This isn't the only reason to set owners through ENS; only 2LD names under .eth have corresponding deeds. I would just leave out this paragraph entirely.

docs/userguide.rst
veox: Hm, agreed. Will change back.

veox: OTOH, what I was trying to achieve here is catching someone who just jumped into the docs and [searched for "transfer name"](https://veox-ens.readthedocs.io/en/doc-ens-setowner-note/search.html?q=transfer+name&check_keywords=yes&area=default), intending to actually transfer full control of it.

There are two results on that page, and the second one shows:

> For example, you can use ens.setOwner to transfer administration of the name to another

Hopefully, they'd still click the link to read the full paragraph, and then realise that `ens.setOwner()` figuring so prominently in the (un!)ellipsized text is not the contract/function they want to call.

docs/userguide.rst
veox: Agreed.

docs/userguide.rst
veox: Basically, I wanted to remove the suggestion that this could be used to transfer to _someone else_.

PR #229, opened 2017-07-29
Added native namehash implementation in Solidity.

Dexaran:
Maybe someone will need it somehow.

### DIFF
--- a/contracts/NameHash.sol
+++ b/contracts/NameHash.sol
@@ -1,0 +1,39 @@
+pragma solidity ^0.4.11;
+
+import "github.com/Arachnid/solidity-stringutils/strings.sol";
+
+
+/**
+ * Namehash algorighm implementation in Solidity.
+ */
+contract NameHash
+ {
+    using strings for *;
+    
+    
+    /**
+     * Native namehash algorighm implementation in Solidity.
+     * @param  _node The string over which the namehash algorithm will be applied.
+     * @return _namehash The result of namehash algorighm applied at _node.
+     */
+    function namehash(string _node) constant returns (bytes32 _namehash)
+    {
+        var node = _node.toSlice();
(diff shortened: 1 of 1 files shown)


Arachnid:
Thanks! A couple of points, though:

 - This needs corresponding unit tests; see the existing tests for examples.
 - This should be a library with inline function, rather than a contract.

veox:
FTR, I've registered `namehash.eth` during the early start.

If a lib like this ever gets merged, and when we have a "community holding" contract for `.eth` names, I'm willing to transfer it.

contracts/NameHash.sol
```diff
@@ -0,0 +1,39 @@
...
+     */
+    function namehash(string _node) constant returns (bytes32 _namehash)
+    {
+        var node = _node.toSlice();
+        bytes32 namehash = 0x0000000000000000000000000000000000000000000000000000000000000000;
+        if(!node.empty())
+        {
+            var delim = ".".toSlice();
+            var parts = new string[](node.count(delim) + 1);
+            for(uint i = 0; i < parts.length; i++)
```
veox: Perhaps `uint256` instead of `uint`, to prevent any possible compiler misrepresentation?..

Arachnid:
> FTR, I've registered namehash.eth during the early start soft launch.

> If a lib like this ever gets merged, and when we have a "community holding" contract for .eth names, I'm willing to transfer it.

Thanks for the offer!

I was looking earlier at implementing UTS46 in the EVM; with this and namehash together, we could have a fully standards compliant implementation that any client could use via local calls, eliminating the risk of nonstandard implementations generating the wrong namehash output.

Unfortunately, the UTS46 lookup tables are _big_ - likely bigger than can fit in a single contract. If anyone's interested in investigating ways of encoding them across multiple contracts, I think that would be a productive avenue of research.

PR #239, opened 2017-10-02
Changed throw to revert() per Solidity 0.4.13

briansoule:
https://github.com/ethereum/solidity/releases/tag/v0.4.13


Arachnid:
Thanks! This should really be `require(...)` for a proper update, though.

PR #242, opened 2017-10-22
changed web3.sha3 to web3.util.sha3 as per web3 1.0.0 change

Brownie79:
web3 1.0.0 update moved web3.sha3 to web3.util.sha3, all new users that do npm install will get the most recent version of web3


Brownie79:
bigger refractor of testcases needed

PR #263, opened 2018-03-07
Initial Refactoring

decanus:
**What does this do:**
* Adds ```public ``` modifiers
* Changes constants to ```view``` / ```pure```
* Changes ```throws``` to ```revert()``` / ```require()```
* Rearranges contracts to use correct function ordering.
* Solidity version updated

**What needs to be done:**
* Add solidity linter
* Split up multi contract files.
* Refactoring where things can be optimized.


contracts/ENSRoot.sol
```diff
@@ -0,0 +1,99 @@
+pragma solidity ^0.4.18;
+
+import './ENS.sol';
+
+/**
+ * The ENS registry contract.
+ */
+contract ENSRoot is ENS {
```
Arachnid: This should definitely be called 'ENSRegistry'; root implies it's only for the root node.

contracts/ENSRoot.sol
```diff
@@ -0,0 +1,99 @@
...
+    modifier only_owner(bytes32 node) {
+        require(records[node].owner == msg.sender);
+        _;
+    }
+
+    /**
+     * @dev Constructs a new ENS registrar.
+     */
+    function ENSRoot() public {
+        records[0].owner = msg.sender;
```
Arachnid: There's a solidity linter error here complaining about left-justified decimal constants (oh no, a left padded zero!) - could you add an explicit cast to `bytes32`?

contracts/HashRegistrarSimplified.sol
```diff
@@ -33,60 +33,60 @@ contract Deed {
...
 
-    function setBalance(uint newValue, bool throwOnFailure) onlyRegistrar onlyActive {
+    function setBalance(uint newValue, bool throwOnFailure) public onlyRegistrar onlyActive {
         // Check if it has enough balance to set the value
-        if (value < newValue) throw;
+        require(value >= newValue);
         value = newValue;
         // Send the difference to the owner
-        if (!owner.send(this.balance - newValue) && throwOnFailure) throw;
+        require(owner.send(this.balance - newValue) && !throwOnFailure);
```
Arachnid: This should be `||`, not `&&`.

contracts/HashRegistrarSimplified.sol
```diff
@@ -347,9 +204,9 @@ contract Registrar {
      *
      * @param sealedBid A sealedBid, created by the shaBid function
      */
-    function newBid(bytes32 sealedBid) payable {
-        if (address(sealedBids[msg.sender][sealedBid]) > 0) throw;
-        if (msg.value < minPrice) throw;
+    function newBid(bytes32 sealedBid) public payable {
+        require(address(sealedBids[msg.sender][sealedBid]) <= 0);
```
Arachnid: Addresses can't be negative.

contracts/HashRegistrarSimplified.sol
```diff
@@ -486,10 +342,12 @@ contract Registrar {
...
      * @param _hash The node to release
      */
-    function releaseDeed(bytes32 _hash) onlyOwner(_hash) {
-        entry h = _entries[_hash];
+    function releaseDeed(bytes32 _hash) public onlyOwner(_hash) {
+        Entry storage h = _entries[_hash];
         Deed deedContract = h.deed;
-        if (now < h.registrationDate + 1 years && ens.owner(rootNode) == address(this)) throw;
+
+        require(now >= h.registrationDate + 1 years);
```
Arachnid: This needs to be a disjunction; we allow releasing if it's been held for a year *or* the registrar has been replaced.

contracts/ENSRoot.sol
Arachnid: If that works, sure.

PR #265, opened 2018-03-21
ENS Tests Fix

decanus:
# What does this PR do?
This PR moves all tests to Truffle, and ensures they properly work.

Things we need to figure out:
- [ ]  Make ens test work with both the lll and sol version


decanus:
@Arachnid Do you know how to make a truffle test work for 2 contracts? I may need to duplicate it

villardez28:
Eth

PR #280, opened 2018-05-14
Make quickstart better

Suhail:
My major goal: Make quickstart easier so more contribute to or adopt ENS. If we can make ENS easy for people to have an immediately gratifying experience, then I believe a higher percentage of them will get involved in the project. The quickstart serves as the second major entry point after the website to start using ENS.

Hey guys, I made the quickstart a more accessible for newbies:
- added some documentation to "connect to Ethereum console"
- updated it to work with rinkeby which is less ingested compared to ropsten
- added some extra things that made me google that came up as obstacles (like unlocking your account)
- emphasized having your node in sync since it was a gotcha for other people having trouble
- added some links to useful resources that people will use anyway like etherscan and geth

Let me know if there's anything you'd like to see changed, broken, or missing.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -4,11 +4,39 @@ Quickstart
 
 Just want to get a name and make it resolve to something? Here's how.
 
-First, download `ensutils-testnet.js`_ to your local machine, and import it into an Ethereum console on a node synced to ropsten or rinkeby:
+First, make sure your client is in **sync with a network** (mainnet, ropsten, rinkeby, etc.). You can use eth.syncing in your Ethereum console to track progress.
+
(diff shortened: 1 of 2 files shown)


docs/quickstart.rst
```diff
@@ -4,11 +4,39 @@ Quickstart
...
+::
+
+        contract address: 0xe7410170f87102df0055eb195163a03b7f2bff4a (line 220)
+        publicResolver address: 0x5d20cf83cb385e06d2f2a892f9322cd4933eacdc (line 1314)
+
+If you need to unlock your account to execute certain transactions:
+
+::
+
+    // Careful as this will record your password in your history.
```
Arachnid: It shouldn't - geth doesn't store any calls to web3.personal in the command history in the console.

Suhail: Yes, learned this later...

PR #282, opened 2018-05-15
FifsRegistrar in ensutils.js has the wrong ABI interface

Suhail:
This should fix this issue: https://github.com/ensdomains/ens/issues/281


Arachnid:
As I said in the issue:

> `expiryTimes` is used because the same ABI is used to access the test registrar. Deleting it here would cause the examples that use the test registrar to fail.

PR #283, opened 2018-05-15
Clarify that the top level TLD is '.'

Suhail:
This greatly confused me for a little while since it wasn't clear how top-level TLDs were really created.

### DIFF
--- a/docs/userguide.rst
+++ b/docs/userguide.rst
@@ -284,6 +284,8 @@ Note the use of `web3.sha3()` instead of `namehash()` when specifying the subdom
 
 The owner of a name can reassign ownership of subdomains at any time, even if they're owned by someone else.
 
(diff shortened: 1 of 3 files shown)


docs/userguide.rst
```diff
@@ -284,6 +284,8 @@ Note the use of `web3.sha3()` instead of `namehash()` when specifying the subdom
 
 The owner of a name can reassign ownership of subdomains at any time, even if they're owned by someone else.
 
+**Note:** When you first create your registry, the top level TLD is '.' (the root register) and the root register creates "subdomains" like 'test' or 'eth'
```
Arachnid: I think you mean 'registrar', not 'register' - and there isn't necessarily a root registrar, it can be owned by a simple account. Perhaps:

When you first create your registry, the only domain is '.' (the root domain) and the root domain creates subdomains like the 'test' and 'eth' top-level domains.

Suhail: Cool, updated

Suhail:
Updated. Should be good now.

PR #286, opened 2018-06-23
Update and rename docs/deploying.rst to  docs/deploying.rst

AUTHAI-bit:
https://github.com/ensdomains/ens/compare/master...AUTHAI-bit:patch-1#diff-a896b57f0e5a70868511a79be3b45f0a


Arachnid:
What?

PR #287, opened 2018-06-23
Create LICENSE

AUTHAI-bit:


LICENSE
```diff
@@ -0,0 +1,25 @@
+BSD 2-Clause License
+
```
AUTHAI-bit: * Author URI: [https://gourl.io](https://gourl.io)

PR #294, opened 2018-08-15
Event emits & Constructors 

JosefJ:
1. Added a subdomain dispensary contract enabling to let anyone register a subdomain at a selected price. 
2. Added event emits
3. Changed constructors to new syntax

The dispensary is a optional wrapper contract enabling anyone to register a subdomain on an owned node in change for fixed price - the owner of the node isn't loosing any setup rights.

### DIFF
--- a/contracts/PublicResolver.sol
+++ b/contracts/PublicResolver.sol
@@ -53,7 +53,7 @@ contract PublicResolver {
      * Constructor.
      * @param ensAddr The ENS registrar contract.
      */
-    function PublicResolver(ENS ensAddr) public {
+    constructor(ENS ensAddr) public {
         ens = ensAddr;
     }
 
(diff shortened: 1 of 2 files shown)


Arachnid:
Thanks for the PR! Would you mind separating out the syntax updates and the 'dispensary' contract into separate PRs?

Regarding the dispensary, I've been cautious about including something like this as a premade component because it's so easy for someone to use it to defraud people, since they retain ultimate control of the domain. We do have the Subdomain Registrar at https://github.com/ensdomains/subdomain-registrar, which prevents this by requiring control of the deed, and I'd need a fairly persuasive argument as to why we should offer a version that doesn't have that requirement.

JosefJ:
I removed the dispensary PR - will first try to figure out if there is really a way to make this happen in one call (delegate call doesn't help as it conflicts with `addr.reverse` node ownership).

PR #297, opened 2018-08-24
Solidity contract maintenance.

hgedia:
- Updating pragma directive to latest compiler
- Supressing event emit warnings
- Using abi.encodePacked for keccak256 (see : https://github.com/ethereum/solidity/issues/3955)

All npm tests pass

### DIFF
--- a/contracts/HashRegistrarSimplified.sol
+++ b/contracts/HashRegistrarSimplified.sol
@@ -431,13 +431,13 @@ contract Registrar {
      * @return The hash of the bid values
      */
     function shaBid(bytes32 hash, address owner, uint value, bytes32 salt) public pure returns (bytes32) {
-        return keccak256(hash, owner, value, salt);
+        return keccak256(abi.encodePacked(hash, owner, value, salt));
     }
 
     function _tryEraseSingleNode(bytes32 label) internal {
         if (ens.owner(rootNode) == address(this)) {
             ens.setSubnodeOwner(rootNode, label, address(this));
-            bytes32 node = keccak256(rootNode, label);
+            bytes32 node = keccak256(abi.encodePacked(rootNode, label));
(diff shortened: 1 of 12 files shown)


contracts/HashRegistrarSimplified.sol
```diff
@@ -431,13 +431,13 @@ contract Registrar {
      * @return The hash of the bid values
      */
     function shaBid(bytes32 hash, address owner, uint value, bytes32 salt) public pure returns (bytes32) {
-        return keccak256(hash, owner, value, salt);
+        return keccak256(abi.encodePacked(hash, owner, value, salt));
```
Arachnid: Pretty sure this should be `abi.encode`; `encodePacked` will pack the address as 160 bits instead of 256.

Arachnid: Can you try a test with an address type, and no variable length (string, bytes) types?

hgedia: pragma solidity ^0.4.24;

    contract echoEncoding{
        event echo(bytes32 echoHash);
        function echoEncoding(){
            address addr1 =  0x314159265dd8dbb310642f98f50c066173c1259b;
            emit echo(keccak256(addr1));
            emit echo(keccak256(abi.encode(addr1)));
            emit echo(keccak256(abi.encodePacked(addr1)));
       }
    } 

Output


    "echoHash": "0x47c21aaccf7858a0bbed6eb11df8b02b87bb4281c9b605ca5302417556015435"
    "echoHash": "0x40a1e975a27fc25470c2a4433cdddd8a359d3114fa0628c5b96f7c954bce754f"
    "echoHash": "0x47c21aaccf7858a0bbed6eb11df8b02b87bb4281c9b605ca5302417556015435"

PR #308, opened 2018-12-27
enhancement/capitalize-constants

### LINKED ISSUE #146: Capitalize constants
http://solidity.readthedocs.io/en/develop/style-guide.html#constants

Constants
Constants should be named with all capital letters with underscores separating words. (for example:MAX_BLOCKS)

(This would also be more in line with conventions in other languages.)

decanus:
closes #146

### DIFF
--- a/contracts/HashRegistrar.sol
+++ b/contracts/HashRegistrar.sol
@@ -31,11 +31,11 @@ contract HashRegistrar is Registrar {
     
     enum Mode { Open, Auction, Owned, Forbidden, Reveal, NotYetAvailable }
 
-    uint32 constant totalAuctionLength = 5 days;
-    uint32 constant revealPeriod = 48 hours;
-    uint32 public constant launchLength = 8 weeks;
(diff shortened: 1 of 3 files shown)


contracts/HashRegistrar.sol
```diff
@@ -31,11 +31,11 @@ contract HashRegistrar is Registrar {
     
     enum Mode { Open, Auction, Owned, Forbidden, Reveal, NotYetAvailable }
 
-    uint32 constant totalAuctionLength = 5 days;
-    uint32 constant revealPeriod = 48 hours;
-    uint32 public constant launchLength = 8 weeks;
+    uint32 constant TOTAL_AUCTION_LENGTH = 5 days;
+    uint32 constant REVEAL_PERIOD = 48 hours;
+    uint32 public constant LAUNCH_LENGTH = 8 weeks;
```
Arachnid: We can't change this, it's public and deployed.

PR #314, opened 2019-02-16
upgrade/test-utils

decanus:


test/TestENS.js
```diff
@@ -1,7 +1,8 @@
-const utils = require('./helpers/Utils.js');
 const namehash = require('eth-ens-namehash');
 const sha3 = require('web3-utils').sha3;
 
+const { exception } = require("@ensdomains/test-utils")
```
Arachnid: Can we use a less generic name than `exception`, which sounds like a builtin?

decanus: you got any suggestions? I couldn't come up with anything better

Ekkarat304:
Thank you

PR #327, opened 2019-03-30
use "then" combinator to map over Promise value

novakov-alexey:
According to **Truffle** spec of `<Contract>.at(address)` method, we have _thenable_ object as return value, so that one needs to map over returned Promise now. https://github.com/trufflesuite/truffle/blob/76f1fb284845210d3cd9983b09e3e7d75f077090/packages/truffle-contract/README.md#mycontractataddress

Closes https://github.com/ensdomains/ens/issues/325


makoto:
cc @Arachnid I don't think we need to publish as new npm version, so I just merge

PR #330, opened 2019-05-04
Add go-ens and Ethereal

mcdee:
go-ens is a Go module that provides support for ENS operations.

Ethereal is a command-line application that includes ENS support.

This adds them to the list of known libraries.


Arachnid:
These are the old docs. The new ones are in ensdomains/docs - can you send a PR to there instead?

PR #333, opened 2019-05-29
Add index file, publish script

jefflau:
- Added tests to npm ignore
- Added a script to build + publish
- Added an index file that exports the JSON contracts


package.json
```diff
@@ -1,12 +1,14 @@
...
   "name": "@ensdomains/ens",
-  "version": "0.3.6",
+  "version": "0.3.8",
   "description": "Implementations for registrars and local resolvers for the Ethereum Name Service",
   "scripts": {
     "test": "truffle test",
     "lint": "solium --dir ./contracts",
-    "prepublishOnly": "truffle compile && truffle networks --clean"
+    "prepublishOnly": "truffle compile && truffle networks --clean",
+    "pub": "yarn prepublishOnly && yarn publish"
```
Arachnid: We don't have a dependency on yarn anywhere else; could you use npm for this instead?

PR #336, opened 2019-12-22
Adding interface to reverse registrar

Amxx:
With solidity 0.6.0 introducing breaking changes, it is not possible for smart contract using solidity 0.6.0 to import contract sources using ^0.5.0 such as those in this repository.

A solution is to, rather then importing the actual source, rely on an interface with pragma >=0.4.24 such as `ENS.sol`. However, there isn't such an interface for the reverse registrar.

This PR solves this by adding such an interface that can be imported by smart contract using ^0.5.0 of ^0.6.0

### DIFF
--- a/contracts/IReverseRegistrar.sol
+++ b/contracts/IReverseRegistrar.sol
@@ -1,0 +1,10 @@
+pragma solidity >=0.4.24;
+
+interface IReverseRegistrar {
+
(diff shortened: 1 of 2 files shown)


Arachnid:
Thanks for this! Are you able to rename to `ReverseRegistrar`/`ReverseRegistrarImplementation` instead, though?

PR #349, opened 2020-03-09
Update Readme to remove references to PublicResolver

wilwade:
PublicResolver and other resolvers are no longer in this package. They have been moved to https://github.com/ensdomains/resolvers so we should not reference them in the Readme except for the link to the new repo.


Arachnid:
Thanks!

PR #399, opened 2022-05-30
npm: Remove unused solc, upgrade package-lock.json

kevinji:
The `solc` package was not actually used as the `solc` binary was being pulled in by Truffle.


TomiOhl:
Would love to have this merged in! The old solc in this package causes security alerts. I know this package is deprecated, but many projects depend on it.

